PR #13, opened 2017-06-24
SSR

### LINKED ISSUE #2: Server-Side Rendering
We probably could do it... 
But it doesn't work as of now because document.createElement isn't available on Node.

image

# Possible Solutions
This might be a good reason to look into [pelo](https://github.com/shuhei/pelo) instead of bel.
Or... I guess the newest version of bel... because that [just exposes pelo now](https://github.com/shama/bel/blob/master/index.js)?

# Tests
Should be fairly easy to test... Actually, if we can point testem to a node instance, that working in and of itself should be a testament to the feature functioning.

### LINKED ISSUE #7: Look into other build tools
Apparently webpack is a little more appropriate for web applications... Let's look into different build tools.

# Possible Solutions
- Rollup

# Tests
All the normal tests should run.. The import might be different.

JRJurman:
## Server Side Rendering
Should allow users to build their app on the server, and then serve up HTML as a string.

## Tram-One Changes
- toString 

## File Changes
- centralized babelrc for builds
- add config for source maps (unverified)
- Node Launcher for Testem, so we can run tests in node.js
- remove asserts from build and dependencies
- two builds, umd and esm
- new build script to build an es module
- conditionalize tram-spec for server runs
- short-circuit browser-specific tests if we're running on a server

## Fixes
- added eslint plugins that were required

## Tests
- TBD

Resolves #2 
Resolves #7 <-- Rollup at its current state is not simple or obvious enough to use for both web bundles and es module builds.


JRJurman:
Not ready quite yet, sanity check proves that exposed build is not functioning as expected.

PR #30, opened 2017-09-22
Hovering

### LINKED ISSUE #29: minidux middleware - the quest for cleaner fetches
## Problem
Currently, fetching data (as you might when interacting with an API) in Tram-One is non-trivial because of the nature of redux (and in turn, minidux). While it's possible to expose adding middleware to minidux, I feel like that goes against the philosophy of Tram-One, which is to include as much as possible on it's own (truly batteries included).

## Solutions
* expose addMiddleware or middleware option in constructor
* bundle a middleware solution (like redux-thunk or redux-saga)
* use a different state management solution
* build something into Tram-One (which would probably need to branch off and become a different state management solution)

I'm personally leaning towards using something other than minidux, because this problem is silly, and frameworks like hyperapp are able to cleanly work around it (probably with their own caveats, but that's neither here nor there).

JRJurman:
resolves #29 

## Summary
This PR uses the [hover-engine](https://github.com/JRJurman/hover-engine) project to give Tram-One better state-management, and allow the project to handle async / chained actions. 

This PR also introduces the [rlite](https://github.com/chrisdavies/rlite) project to give Tram-One better routing (and query params).

**This introduces a breaking change to Tram-One, pushing the version to 2.0.0**

## New API

There is now an `addActions` function, to replace the `addReducers` function, this should make building state in Tram-One simpler and more predictable.

### before
```javascript
app.addReducer('counter', (state, action) => {
  switch(action.type) {
    case 'UP':
      return state+1
    default:
      return state
  }, 0)
```

### now
```javascript
app.addActions({counter: {
  init: () => 0,
  up: (count) => count + 1
}
```

Views now take in store, actions, and params.

### before
```javascript
page = (state) => html`
  <div>
    <button onclick=${() => state.dispatch({type: 'ACTION_NAME'})}>Click Me</button>
    <div>${state.userId}</div>
  </div>
`
```

### now
```javascript
page = (store, actions, params) => html`
  <div>
    <button onclick=${() => actions.actionName()}>Click Me</button>
    <div>${params.userId}</div>
  </div>
`
```

## Changes
- run lint as part of circleci run
- switch to xo instead of eslintrc
- update README dependencies
- include size in README
- update README for new API
- update package.json destinations to `pkg.main` (for `umd`) and `pkg.module` (for `esm`)
- use `babel-preset-env` in rollup
- removed babel plugins and presets in dev-dependencies
- re-order rollup dependencies to be alphabetical
- change test-example to use jasmine instead of testem
- changed lots of examples to use new API
- updated specs for new API
- new `package-lock.json` file
- update most dependencies 
- rename config files
- update `tram-one.js` to be AWESOME!


ethanjurman:
Super approved

PR #38, opened 2017-10-07
fix test-example: update to use new api

JRJurman:
## Summary
The test example wasn't updated to use the new `params` parameter 

## Changes
- Update the page and spec to use the new parameters


JRJurman:
Not merged in because CircleCI tests aren't running, will merge in once they are good.

PR #49, opened 2017-10-15
Example Added : Tram-One Image Carousel

KhushilMistry:


examples/image-carousel/reducers/ImageCarouselActions.js
```diff
@@ -0,0 +1,46 @@
+module.exports = {
+  init: () => Object({
+    tiles: [{
+      image: 'https://www.w3schools.com/howto/img_nature_wide.jpg',
+      text: 'Nature and sunrise',
+      alt: 'Nature and sunrise'
+    }, {
```
ethanjurman: I would move tiles out of the actions, so that way you can just have `selected` be the only thing you are tracking (since you are not altering tiles at all).
You wouldn't have length in `onSelectImage`, `onNextImage`, etc... so you would have to move that logic into the viewing components. @JRJurman your thoughts?

examples/image-carousel/index.js
```diff
@@ -0,0 +1,44 @@
...
+  }
+
+  const onNextImage = () => {
+    actions.onNextImage();
+  }
+
+  return html`
+    <div>
+      <h1 style=${textCenter}> Tram-One Image Carousel 🚋 </h1>
+      <Image tiles=${store.data.tiles} selected=${store.data.selected} onNextImage=${onNextImage} onPreviousSelect=${onPreviousSelect}></Image>
```
ethanjurman: Assuming you have Tram v3 (just updated today), you can actually self close this :D 
So you can just do 
```js
<Image tiles=${store.data.tiles} selected=${store.data.selected} onNextImage=${onNextImage} onPreviousSelect=${onPreviousSelect} />
```

examples/image-carousel/elements/Image.js
```diff
@@ -0,0 +1,52 @@
...
+
+  const img = attrs.tiles[attrs.selected];
+
+  return html`
+    <div style=${divStyle}>   
+        <img style=${imgStyle} src=${img.image} alt=${img.alt}/>
+        <h3 style=${textCenter}>${img.text}</h3>
+      <h3>
+        <span onclick=${attrs.onPreviousSelect} style=${textLeft}>Previous</span>
+        <span onclick=${attrs.onNextImage} style=${textRight}>Next</span>
```
ethanjurman: Is there a reason you are not using a button here? 
If you keep these as spans make sure that you use  `user-select: none;` in your style as there is a weird highlight text issue if you click the span multiple times.

examples/image-carousel/index.js
ethanjurman: Same with `ImageDeck` below

examples/image-carousel/elements/ImageDeck.js
```diff
@@ -0,0 +1,30 @@
...
+  width : 100%;
+  border : 2px black solid;
+  overflow : scroll;
+  display : flex;
+`
+
+module.exports = (attrs, children) => {
+  const renderImageTile = (value,index) => {
+    return html`
+      <ImageTile value=${value} index=${index} onSelect=${attrs.onSelect(index)}></ImageTile>
```
JRJurman: This can be a self-closing tag now! 
```
<ImageTile value=${value} index=${index} onSelect=${attrs.onSelect(index)}></ImageTile>
```

examples/image-carousel/elements/Image.js
```diff
@@ -0,0 +1,52 @@
...
+
+const html = Tram.html()
+
+const divStyle = `
+  width : 100%;
+`
+
+const imgStyle = `
+  width : 100%;
+  height : 400px;
```
JRJurman: The fixed height actually is stretching the images, you can just remove this line.

examples/image-carousel/elements/Image.js
```diff
@@ -0,0 +1,52 @@
...
+`
+
+const textLeft = `
+  float : left;
+  padding : 5px 7px;
+  background-color : black;
+  color : white;
+  cursor : pointer;
+  margin-bottom : 10px;
+`
```
JRJurman: a lot of this style is unnecessary, this really could be boiled down to
```javascript
const textLeft = `
  float : left;
  cursor : pointer;
`
```

examples/image-carousel/elements/Image.js
```diff
@@ -0,0 +1,52 @@
...
+
+const textRight = `
+  float : right;
+  padding : 5px 7px;
+  background-color : black;
+  color : white;
+  cursor : pointer;
+  margin-bottom ; 10px;
+`
```
JRJurman: same as above
```javascript
const textRight = `
  float : right;
  cursor : pointer;
`
```

examples/image-carousel/elements/Image.js
JRJurman: Buttons can have odd styling, so I'm actually okay with a span here. `user-select: none;` would be cool, but I'm not that worried about it.

examples/image-carousel/elements/ImageTile.js
```diff
@@ -0,0 +1,23 @@
...
+const divStyle = `
+  padding : 20px;
+  display: flex;
+  cursor : pointer;
+`
+
+module.exports = (attrs, children) => {
+  return html`
+    <div style=${divStyle}>
+      <abbr title=${attrs.value.alt}><img style=${imgStyle} src=${attrs.value.image} alt=${attrs.value.alt} onclick=${attrs.onSelect}/></abbr>
```
JRJurman: Huh, I never knew of the `<abbr>` tag before... That being said, it's probably not necessary for this example.

examples/image-carousel/reducers/ImageCarouselActions.js
JRJurman: I agree, if it's possible, I would definitely recommend only storing logic that you anticipate on changing

JRJurman: If you want to, you could store the number of tiles in here, just so that there aren't as many changes required for the actions below.

KhushilMistry:
Thanks for quick response. I will update all these changes.

KhushilMistry:
Hey, I have added required changes. Can you look at it ?

KhushilMistry:
Thanks for merging this !!

Issue #50, opened 2017-10-17
Routing Example


JRJurman:
## Summary
It would be cool if one of the example apps in the tram-one examples folder had multiple pages and showed how we keep state between pages. It should be more than just the current `using-routes` example (i.e. it should be an app that is intrinsically useful, as is the case with all the examples).

PR #53, opened 2017-10-18
insert example using-routes-todo-app

davidsonsns:
based on the Point-Cards example. I used localStorage to store the state

issue #50


examples/using-routes-todo-app/index.js
```diff
@@ -0,0 +1,82 @@
...
+}
+
+app.addActions({todos: TodoActions})
+app.addRoute('/', home)
+app.addRoute('/todo', todo)
+app.addRoute('/results', results)
+
+app.engine.addListener((store) => {
+  localStorage.setItem('todos', JSON.stringify(store.todos))
+})
```
JRJurman: Nice use of `addListener`!

That being said, we don't need this if the links do `window.history.pushState({}, '', href)`
It means that page-refreshes will clear all the information, but for right now, I think I'd rather do that then use the engine's addListener function.

JRJurman: Making this change would also require all the `<a>` tags to be rewritten as tags with a `onclick` that does the `pushState`

examples/using-routes-todo-app/index.js
```diff
@@ -0,0 +1,82 @@
...
+  const onCompleteTodo = (index) => () => {
+    actions.completeTodo(index)
+  }
+
+  return html`
+    <div>
+      <div class="center">
+        <h1> Tram-One Todos 🚋 </h1>
+        <a href="/">HOME</a> | <a href="/results">RESULTS</a>
+      </div>
```
JRJurman: to reduce how many times we write this, lets make a custom element which is the header with all the links in it (even if it a bit redundant)

davidsonsns:
I'm glad to hear that you liked it!

your review was really good, I'll do it.

I did not know of the existence of `pushState` in `history`, thanks for the explanation

examples/using-routes-todo-app/index.js
davidsonsns: I did not understand what you meant. so I should remove `addListener` and use `pushState`?

examples/using-routes-todo-app/elements/HeaderPage.js
```diff
@@ -0,0 +1,13 @@
...
+
+module.exports = ({title, subtitle = undefined}, children) => {
+    const contentSubtitle = subtitle ? html`<h2> ${subtitle} </h2>` : null
+  return html`
+    <div>
+      <h1> ${title} 🚋 </h1>
+      ${contentSubtitle}
+    </div>
+  `
+}
```
JRJurman: oh ho! I like this 👍

examples/using-routes-todo-app/index.js
JRJurman: I'm actually not 100% sold that we should change to this, so I'm approving the PR for now. We may change it in the future, but for now, this example is really good, so I'd rather get it in, and clean it up later if-need-be.

davidsonsns: @JRJurman  I think if we leave both options would be interesting. maybe this same example implementing `pushState`

I really enjoyed the experience, I will continue to contribute with certainty

PR #54, opened 2017-10-18
Expose Hover-Engine's addListener

### LINKED ISSUE #33: Expose Hover-Engine's addListener
## Summary
It would be useful to have a chrome-extension to listen for state changes from hover-engine (similar to the redux chrome-extension). In order to even start that work, we need to be able to add listeners to tram-one state changes.

This should be pretty similar to Tram-One's addActions, where we just call `this.engine.addListener`
https://github.com/Tram-One/tram-one/blob/b82d2feda534031f85b97124d4cb7d2bc650eeeb/tram-one.js#L23-L32

Docs for `addListener`
https://github.com/Tram-One/hover-engine#addlistenerlistener

## Todo
For this ticket we need the following things:
* a new function `addListener`, it should take in a listener and return `this`
* there should be an assert that the listener is a function
* it should call `this.engine.addListener` with the passed in listener

We also want the following test:
* on calling an action, the custom listener should be called (you don't have to check with what, just that it was)

### Example Custom Page
You'll need a custom page that triggers an action once when it loads (and only once when it loads)
```javascript
const pageWithAction = (store, actions) => { 
  if (store.switch) { actions.toggle() } 
  return html`<div>${store.switch}</div>`
}

const switchAction = {
  init: () => true
  toggle: () => false
}
```

KhushilMistry:
Resolves #33


tram-one.js
```diff
@@ -31,6 +31,17 @@ class Tram {
     return this
   }
 
+  addListener(listener) {
+    assert.equal(
+      typeof listener, 'object',
+      'Tram-One: listener should be { store-key: { listener-name: listener-function } }'
+    )
```
JRJurman: the assert should be of type function, you can look at the addRoute method for an example of that.

tests/specs/tram-spec.js
```diff
@@ -71,6 +72,18 @@ const tests = (Tram) => describe('Tram', () => {
     })
   })
 
+  describe('addListener', () => {
+    it('should call single listener', () => {
+      const spy = jasmine.createSpy('listener')
+      const app = new Tram()
+      app.addActions({counter: counterActions})
+      app.engine.addListener(spy)
```
JRJurman: we should call `app.addListener(spy)` since that is the function we are exposing.

tests/specs/tram-spec.js
```diff
@@ -71,6 +72,18 @@ const tests = (Tram) => describe('Tram', () => {
...
+      app.engine.notifyListeners()
```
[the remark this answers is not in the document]
KhushilMistry:
```
app.addActions({counter: counterActions})
app.addListener(spy)
app.engine.actions.add()

expect(spy).toHaveBeenCalled()
```

Like this ?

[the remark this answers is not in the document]
JRJurman: Yep! That should should be perfect.  I take back what I said before, we probably want to test that the spy was called *with* the newest version of the store, not just that it was called at all.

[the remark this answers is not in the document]
KhushilMistry: How can we test that ?

tram-one.js
```diff
@@ -31,6 +31,13 @@ class Tram {
     return this
   }
 
+  addListener(listener) {
+    assert.equal(typeof listener, 'function', 'Tram-One: page should be a function')
```
JRJurman: this should say `Tram-One: listener should be a function`

tests/specs/tram-spec.js
```diff
@@ -71,6 +72,18 @@ const tests = (Tram) => describe('Tram', () => {
     })
   })
 
+  describe('addListener', () => {
+    it('should call single listener', () => {
+      const spy = jasmine.createSpy('listener')
+      const app = new Tram()
+      app.addActions({counter: counterActions})
+      app.addListener(spy)
+      app.engine.actions.add()
```
JRJurman: instead of `app.engine.actions.add()` we want the listener to be called from Tram-One (right now this really only tests that hover-engine works).

Instead, we should have a page that triggers an action when the app is started.

tests/specs/tram-spec.js
```diff
@@ -30,6 +30,7 @@ const tests = (Tram) => describe('Tram', () => {
   const counterPage = (store) => Tram.html()`${store.counter}`
   const queryableCounterPage = (store) => Tram.html()`<div id="tram_container">${store.counter}</div>`
 
+
```
JRJurman: probably don't need this newline?

tests/specs/tram-spec.js
KhushilMistry:
```
app.addActions({actions: switchAction})
app.addListener(spy)
pageWithAction(app.engine.store, app.engine.actions)

expect(spy).toHaveBeenCalled()
```
```
const pageWithAction = (store, actions) => {
    const html = Tram.html()
    if (store) { actions.toggle() }
    return html`<div>${store}</div>`
}
const switchAction = {
    init: () => true,
    toggle: () => false
}
```
Something like this ?

JRJurman: Yeah! But, instead of `pageWithAction(app.engine.store, app.engine.actions)` we'll want to do the same thing we do with the other tests, and actually call `app.start`

JRJurman:
@KhushilMistry what's the status of this PR? if you don't want to continue the PR, that's fine, there's no pressure. If you want help, there's a slack now (join link in the README), and maybe we can work on closing this out.

JRJurman:
Issue was resolved in similar PR #65

PR #55, opened 2017-10-18
generate build-size badges 

### LINKED ISSUE #52: Badge for Build Size
## Summary
It would be awesome if we could have a custom `umd` and `module` size badges.

## Possible Implementation
I'm not 100% sure what the best way to do this is, but I believe one way we could go about this is by using the [shields project](https://github.com/badges/shields) (which is used for the other badges anyways) as a dependency. On builds (or post-build) we read the size of the distributables, and generate the images using the library. The github readme could then point to these local assets for the build size badges.

thatbraxguy:
Added postbuild script to generate build size badges

Resolves #52


README.md
```diff
@@ -4,6 +4,8 @@
   <a href="https://www.npmjs.com/package/tram-one"><img src="https://img.shields.io/npm/dm/tram-one.svg" alt="Downloads"></a>
   <a href="https://www.npmjs.com/package/tram-one"><img src="https://img.shields.io/npm/v/tram-one.svg" alt="Version"></a>
   <a href="https://www.npmjs.com/package/tram-one"><img src="https://img.shields.io/npm/l/tram-one.svg" alt="License"></a>
+  <img src="./docs/images/esm-size.svg" alt="License">
+  <img src="./docs/images/umd-size.svg" alt="License">
```
JRJurman: should have different alt

dev-scripts/build-size-badges.js
```diff
@@ -0,0 +1,42 @@
...
+        if (err) {
+            process.stdout.write('Error generating build size badge', err);
+        } else {
+            save(svg, label);
+        }
+    });
+}
+
+generateBadge('esm-size', getSize('tram-one.esm.js'));
+generateBadge('umd-size', getSize('tram-one.esm.js'));
```
JRJurman: this should be `tram-one.umd.js`

PR #69, opened 2018-01-29
Use Belit and Domino

JRJurman:
## Summary
This update includes the updated bel-create-element library, and substitutes min-document for domino. This should make server-side rendering more consistent with client-side rendering.

## Misc Changes
- Use new and improved the library formally known as bel-create-element : **belit**!
- Use domino (which belit uses) to generate virtual document objects on the server
- Exclude domino from umd builds (which we should have done with min-document)

#### `Tram.dom(namespace, registry)`
The same as `Tram.html` but takes in a namespace, so you can create xml of different types.

#### `Tram.svg(registry)`
The same as `Tram.html`, shorthand for creating svg elements. (required now for elements which are svg)

## Todo
- [x] create tests for `Tram.dom`
- [x] create tests for `Tram.svg`
- [x] create svg example


JRJurman:
Don't know why but the test keep on failing when run with cache. When they are rebuilt without cache they pass... If this continues to be an issue, one of these commits is to blame...

docs/images/esm.svg
```diff
@@ -1 +1 @@
...
\ No newline at end of file
+<svg xmlns="http://www.w3.org/2000/svg" width="82" height="20"><linearGradient id="b" x2="0" y2="100%"><stop offset="0" stop-color="#bbb" stop-opacity=".1"/><stop offset="1" stop-opacity=".1"/></linearGradient><mask id="a"><rect width="82" height="20" rx="3" fill="#fff"/></mask><g mask="url(#a)"><path fill="#555" d="M0 0h33v20H0z"/><path fill="#e05d44" d="M33 0h49v20H33z"/><path fill="url(#b)" d="M0 0h82v20H0z"/></g><g fill="#fff" text-anchor="middle" font-family="DejaVu Sans,Verdana,Geneva,sans-serif" font-size="11"><text x="16.5" y="15" fill="#010101" fill-opacity=".3">esm</text><text x="16.5" y="14">esm</text><text x="56.5" y="15" fill="#010101" fill-opacity=".3">3.65 kB</text><text x="56.5" y="14">3.65 kB</text></g></svg>
```
chtinahow: Are you following eslint? Cause you should do the new line thing if you are

docs/images/umd.svg
```diff
@@ -1 +1 @@
...
\ No newline at end of file
+<svg xmlns="http://www.w3.org/2000/svg" width="88" height="20"><linearGradient id="b" x2="0" y2="100%"><stop offset="0" stop-color="#bbb" stop-opacity=".1"/><stop offset="1" stop-opacity=".1"/></linearGradient><mask id="a"><rect width="88" height="20" rx="3" fill="#fff"/></mask><g mask="url(#a)"><path fill="#555" d="M0 0h33v20H0z"/><path fill="#e05d44" d="M33 0h55v20H33z"/><path fill="url(#b)" d="M0 0h88v20H0z"/></g><g fill="#fff" text-anchor="middle" font-family="DejaVu Sans,Verdana,Geneva,sans-serif" font-size="11"><text x="16.5" y="15" fill="#010101" fill-opacity=".3">umd</text><text x="16.5" y="14">umd</text><text x="59.5" y="15" fill="#010101" fill-opacity=".3">14.91 kB</text><text x="59.5" y="14">14.91 kB</text></g></svg>
```
chtinahow: Same thing here as above

tests/specs/tram-spec.js
```diff
@@ -128,8 +121,8 @@ const tests = (Tram) => describe('Tram', () => {
         .addRoute('/good', successPage)
         .addRoute('/bad', errorPage)
 
-      expect(app.toString('/good')).toEqual(stringify(successPage()))
-      expect(app.toString('/bad')).toEqual(stringify(errorPage()))
+      expect(app.toString('/good')).toEqual(successPage().outerHTML)
+      expect(app.toString('/bad')).toEqual(errorPage().outerHTML)
```
chtinahow: Why do you not check the 404 route here as well?

tests/specs/tram-spec.js
```diff
@@ -310,25 +303,76 @@ const tests = (Tram) => describe('Tram', () => {
...
+      const tramTree = Tram.dom()`<div><span></span></div>`
       const docTree = document.createElement('div')
       docTree.appendChild(document.createElement('span'))
       expect(tramTree.outerHTML).toBe(docTree.outerHTML)
     })
 
     it('should take in a registry', () => {
-      const foo = () => Tram.html()`<div><span></span></div>`
-      const tramTree = Tram.html({foo})`<foo></foo>`
+      const foo = () => Tram.dom()`<div><span></span></div>`
```
chtinahow: Why are we using foo for the test? Could this be named something more useful for the registry?

docs/images/esm.svg
JRJurman: this is auto generated.
Having it split into multiple lines would be nice and make future diffs significantly smaller (we should just be updating the size difference).

Either way, I'll make an issue to look into it later.

docs/images/umd.svg
JRJurman: same answer as above

docs/images/esm.svg
JRJurman: https://github.com/Tram-One/tram-one/issues/70

tests/specs/tram-spec.js
JRJurman: the test here is that `.addRoute` returns the Tram One object, and can be called after itself. 

The default 404 route is already tested here: 
https://github.com/Tram-One/tram-one/blob/c0c1323455e83ff70ac550a303adabb57df3e6aa/tests/specs/tram-spec.js#L29-L34

I guess it might be useful to check that a route isn't overwritten by the addition of other routes, but that would be logic to test in `rlite`

tests/specs/tram-spec.js
JRJurman: good point, will fix!

JRJurman:
SVG Clock example is now on this branch
screenshot 2018-02-05 at 11 04 28 am

tests/specs/tram-spec.js
JRJurman: it has been done⚡️

README.md
```diff
@@ -232,7 +232,7 @@ Tram-One has a simple interface to help build your web app.
 
 ### `Tram.html([registry])`
 _Reference: [hyperx](https://github.com/substack/hyperx),
-[bel-create-element](https://github.com/JRJurman/bel-create-element),
+[belit](https://github.com/JRJurman/belit),
```
ethanjurman: Shouldn't this be `Tram-One/belit`?

JRJurman: yes, yes it should!... will fix!

JRJurman: it has been done ⚡️ 
as well as all the other JRJurman references

examples/clock-example/elements/hand.js
```diff
@@ -0,0 +1,36 @@
...
+const handStyle = (degrees) => `
+  transform-origin: center;
+  transform: rotate(${degrees}deg);
+`
+
+const degreesFromAttrs = (attrs) => {
+  return 360 * (attrs['value'] / attrs['max-value'])
+}
+
+const halfHand = svg`
```
ethanjurman: could you label this as halfHand definition? this might be a tad confusing for people that don't use svgs a lot.
Or maybe you could move this to an area where all the defs can live.

JRJurman: I've been thinking back and forth about moving this definition out, and potentially just creating one instance of the def in `clock.js`, but then this element becomes dependent on an outside dependency, which feels really weird.

I can rename it to be something more meaningful.

PR #71, opened 2018-02-09
Nested Routes

### LINKED ISSUE #32: Nested Routing
## Summary
React router is really cool, and lets you easily describe wrapping page elements around components. 

```javascript
<Route path="/" component={PageChrome}>
  <Route path=":user" component={UserPage} />
</Route>
```

Doing this in Tram-One would require a possibly breaking change to `addRoute`, and definitely a breaking change to what properties top-level / low-level view take in. This might be a sign that there shouldn't be a distinction between top and low level views.

### LINKED ISSUE #72: CircleCI failing on PR builds
## Summary
For some reason, CircleCI always fails unless repeated without cache. This happened in the PR for #69, but it could be something external as well.

JRJurman:
resolves #32 

## Summary
Other frameworks (namely react-router) allows you to describe wrapping components that are used based on routes. Namely, you would have a main page, which always has a header and nav bar, and then you swap out the internals based on the route. This PR introduces functionality to allow for that.

### Top Level Page Components
Add new parameter to top level components, `child`. This component is rendered by the time it is used, and is passed in all the same properties as a top level page component (store, actions, params, child).
```javascript
const page = (store, actions, params, child) => html`
  <div>
    <h1>Title</h1>
    ${child}
  </div>
`
```

### route
`route` is a static method which is a shorthand for creating the objects used in the `addRoute` method (detailed below). `route` takes in `path`, `component`, and `subroutes`, mimicking the interface used in `addRoute`.

### addRoute
`addRoute` now takes in a new parameter `subroutes`. This is how the children mentioned above are decided, based on the url path resolved.
```javascript
app.addRoute('/', chrome, [
  route('alpha', require('./pages/alpha')),
  route('beta', require('./pages/beta')),
  route('gamma', require('./pages/gamma'), [
    route('/one', require('./pages/gamma-one')),
    route('/two', require('./pages/gamma-two'))
  ])
])
``` 
In this example (pulled from the `using-nested-routes` example in the project, will generate a `chrome` element at the top level, and inside of that `alpha`, `beta`, or `gamma`. The `gamma` page has it's own subroutes, `/one` and `/two`, which will render inside the Gamma page. 

## Misc Changes
- Fixed #72 

## Todo
- [x] Tests
- [x] Update PR description
- [x] README update


README.md
```diff
@@ -442,14 +449,18 @@ app.addActions({votes: voteActions})
 
 </details>
 
-### `app.addRoute(path, page)`
+### `app.addRoute(path, page[, subroutes])`
```
chtinahow: Why is the comma added within the Array?

README.md
```diff
@@ -487,7 +498,51 @@ app.addRoute('/404', noPage)
...
+Nested Routes Example:
+</summary>
+
+```js
+/* index.js */
+const app = new Tram()
+const html = Tram.html()
+const route = Tram.route()
+
+const homePage = (store, actions, params, child) => {
```
chtinahow: Are all of these parameters being used?

examples/using-nested-routes/elements/selection.js
```diff
@@ -0,0 +1,8 @@
+const Tram = require('../../../tram-one')
+const html = Tram.html()
+
+module.exports = (attrs, children) => html`
+  <div style="display: flex; justify-content: space-around; padding: 0.5em;">
```
chtinahow: I personally would prefer this to be pulled out into its own little styling area but just my opinion

README.md
JRJurman: this is the syntax used when a parameter is optional. It's not the universal standard, but I've seen it used in man pages. I'm open to other suggestions though.

README.md
JRJurman: no, but the last one is, which means they all have to be written out.

examples/using-nested-routes/elements/selection.js
JRJurman: eh, this element isn't really anything except that styling. At that point, I would be tempted to just have the CSS string as a variable in the parent page...

README.md
chtinahow: You could also use _ correct?

examples/using-nested-routes/elements/selection.js
chtinahow: Then do that?

README.md
JRJurman: Yes, but I would argue as an example it's good to repeat the API.

examples/using-nested-routes/elements/selection.js
JRJurman: eh, I like this as another example of using registered elements.

tram-one.js
```diff
@@ -40,10 +40,21 @@ class Tram {
...
 
-  addRoute(path, page) {
+  addRoute(path, page, subroutes) {
     assert.equal(typeof path, 'string', 'Tram-One: path should be a string')
     assert.equal(typeof page, 'function', 'Tram-One: page should be a function')
 
+    if (subroutes) {
+      subroutes.forEach(subroute => {
+        const newPath = path + subroute.path
+        const newPage = (store, actions, params, child) => {
```
ethanjurman: Would make this function name clear that it is *doing* something. Good rule of thumb is always have a verb with function names, especially when using `const`. This makes it easier to reason what its purpose is.

ethanjurman: Maybe `makeNewPage`

PR #77, opened 2018-02-13
Fixes! 🚒

### LINKED ISSUE #72: CircleCI failing on PR builds

### LINKED ISSUE #75: [BUG] BabelHelpers Missing
## Summary
There is currently a bug in version 4.0.0, where when trying to addActions, an error is thrown because BabelHelpers is not defined.

This issue is to track the progress of the work.

JRJurman:
## Summary
In a blind attempt to get tests working, I did extra magic (added babel external helpers) in the webpack test builder, that was not replicated in the rollup builds (100% my bad).

### bublé 
Getting the babel external helpers working in the rollup build was non-trivial, combined with the frustration of all the babel configuration nonsense that the project has had to endure up until this point, I'm taking a gamble and using a no-configuration tool, bublé. It is a gamble, and should we decide to switch back to babel, we'll have some pretty good ideas on why then and there.

### Polyfill
In order to have `Object.assign` replaced, there are a ton of substitutions that can be made with babel, but sadly nothing that really works with bublé. A best case solution would be a rollup plugin that adds it to the top, but for now, we're doing a require (which yes, get's pulled into the esm build, but we can fix that later).

### Fixes
- fix for missing babelHelpers (resolves #75)
- fix failing circleCI builds (resolves #72)

### New Bugs
- Tests pass locally, but are failing in CircleCI #80
- Polyfill is included in esm build (where it is 100% not required) #79 

### Regressions?
Because this issue was cause by a global variable being available in the tests and not in the actual build, I'm not 100% sure how we could go about stopping this in future builds. I willing to explore options here (probably in a different PR).


JRJurman:
I am merging this in now so that we have a working project in master.  
I believe I have created all the appropriate follow-up tickets.  
Should other regressions emerge, we can make tickets as those appear.

PR #99, opened 2018-05-28
Server Side Rendering & Hydration Example

### LINKED ISSUE #35: Example Hydrate Server-Side-Rendered App
## Summary
There is no documentation or obvious way to take an app that has been generated using SSR, and add state-management (aka hydration).

It would be valuable to determine if and how this would work in Tram-One. For now an example that shows this in action would be awesome!

JRJurman:
## Summary

Previously, for our Server Side Rendering (SSR) example, we could only render the page without javascript. The example has been updated to include a script tag pointing to a webpack build, which should "hydrate" the app with javascript on page load.

Resolves #35

### DIFF
--- a/examples/ssr-example/server.js
+++ b/examples/ssr-example/server.js
@@ -18,7 +18,8 @@ server.use(express.static('dist'))
 // serve the home page, with no bundle
 // because it needs no hydration
 server.get('/', (req, res) => {
-  res.send(app.toString('/'))
+  const page = app.toString('/')
+  res.send(page)
 })
 
 // serve the number page with a bundle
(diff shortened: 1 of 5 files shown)


examples/ssr-example/server.js
```diff
@@ -4,15 +4,32 @@ const express = require('express')
...
+  <script src="${src}"></script>
+`
+
+server.use(express.static('dist'))
+
+// serve the home page, with no bundle
+// because it needs no hydration
 server.get('/', (req, res) => {
-  res.send(app.toString('/'))
+  const page = app.toString('/')
```
chtinahow: Why did you decide to break this out into two lines?

JRJurman: It's not required for the home page, but is required for the other example, so I wanted to keep them consistent...

Conversely, we could have both call the `pageWithBundle` function, but that would over-complicate what is required for a page that needs no hydration

PR #103, opened 2018-09-29
Docs and webEngine option

### LINKED ISSUE #102: Document methods that can be consumed
## Summary

Right now IDEs do not provide details when typing out methods. Perhaps documenting these methods could make it easier.

JRJurman:
Resolves #102 

## JSDoc

This PR adds JSDoc syntax enough so that users with IDEs should get type hints and descriptions on calling methods. I'm still reserving most of the documentation to the website (since that offers infinitely more clarity).

## `options.webEngine`

This PR also adds a `webEngine` option that can be passed into the Tram-One constructor. This option allows users to pass in an object that will have the latest store and actions object.

With this, we can enable "connected" components, in a similar way to redux connected components, that don't require values or actions to be passed several tags down.

In the following example (found under `examples/using-webEngine`) we've passed in `{webEngine: window.engine}` into the app constructor. Rather than pulling from attrs, we can now pull directly from `window.engine`.

```javascript
const Tram = require('../../tram-one')
const html = Tram.html()

module.exports = (attrs, children) => {
  const {store, actions} = window.engine

  const incrementCount = () => {
    console.log(`I've been clicked!`)
    actions.click()
  }

  return html`
    <button onclick=${incrementCount}>
      This button has been clicked ${store.counter} times!
    </button>
  `
}
```

You could also build out a `connect` method if you wanted, that returns this method, with the attrs filled in by window.engine (this would effectively mimic the mapStateToProps and connect method that redux provides).

```javascript
const Tram = require('../../tram-one')
const html = Tram.html()

// this is a similar example of counter-button, but using
// normal attrs, and exposing both a connected, and non-connected
// variant (useful for testing and sharing across projects)

const counterButton = (attrs, children) => {
  const incrementCount = () => {
    console.log(`I've been clicked!`)
    attrs.onclick()
  }

  return html`
    <button onclick=${incrementCount}>
      This button has been clicked ${attrs.counter} times!
    </button>
  `
}

module.exports = (attrs, children) => {
  const {store, actions} = window.engine
  const connectedAttrs = Object.assign({}, attrs, {counter: store.counter, onclick: actions.click})
  return counterButton(connectedAttrs, children)
}

module.exports.counterButton = counterButton
```


docs/images/esm.svg
```diff
@@ -14,7 +14,7 @@
   <g fill="#fff" text-anchor="middle" font-family="DejaVu Sans,Verdana,Geneva,sans-serif" font-size="11">
     <text x="16.5" y="15" fill="#010101" fill-opacity=".3">esm</text>
     <text x="16.5" y="14">esm</text>
-    <text x="59" y="15" fill="#010101" fill-opacity=".3">4.31 kB</text>
-    <text x="59" y="14">4.31 kB</text>
+    <text x="59" y="15" fill="#010101" fill-opacity=".3">9.48 kB</text>
+    <text x="59" y="14">9.48 kB</text>
```
JRJurman: this doubled because of the JSDocs, but shouldn't matter, because this is the server (check the UMD file for web impact)

docs/images/umd.svg
```diff
@@ -14,7 +14,7 @@
   <g fill="#fff" text-anchor="middle" font-family="DejaVu Sans,Verdana,Geneva,sans-serif" font-size="11">
     <text x="16.5" y="15" fill="#010101" fill-opacity=".3">umd</text>
     <text x="16.5" y="14">umd</text>
-    <text x="59" y="15" fill="#010101" fill-opacity=".3">11.66 kB</text>
-    <text x="59" y="14">11.66 kB</text>
+    <text x="59" y="15" fill="#010101" fill-opacity=".3">11.87 kB</text>
+    <text x="59" y="14">11.87 kB</text>
```
JRJurman: a little larger, but not significantly so..

tram-one.js
```diff
@@ -58,66 +103,144 @@ class Tram {
...
+   * generate a DOM Node Tree based on a path and a store
+   * - for complete documentation, please refer to http://tram-one.io/#app-toNode
+   *
+   * @param {string} pathName
+   * @param {object} [store]
+   * @param {object} [actions]
+   *
+   * @return {HTMLDivElement}
+   */
+  toNode(pathName, store, actions) {
```
ethanjurman: Thoughts of using optional parameters here? Obviously this wouldn't be equivalent, but I think if someone wants to pass in a false-y store and actions you shouldn't be overriding them.

ethanjurman:
Minor bump?

JRJurman:
Most probably

JRJurman:
Not yet... But it's something I should make a ticket for

examples/using-webEngine/index.html
```diff
@@ -0,0 +1,10 @@
+<html>
```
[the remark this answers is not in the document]
JRJurman: Not yet, but I should make a ticket for it

tram-one.js
JRJurman: Actually I would love to changes these all up to use parameters deconstruction... However I don't know if I could do type checking then... Or which is better

ethanjurman: You could still do optional parameters with objects. And probably still do the asserts your doing above.
```js
toNode({pathName, store=this.engine.store, actions=this.engine.actions}) {
```

Or you could always do a typeof check for objects and support objects and individual parameters

PR #106, opened 2018-10-03
Allow Default Subroutes

JRJurman:
## Summary

In the past, it was non trivial to have a default child for a route when using subroutes. This is because if you defined a child at `/` and a parent at `/`, the child would be assigned in the router first, and then replaced by the parent definition.
```javascript
app.addRoute('/', require('./pages/wrapper'), [ // assigned second
  route('/', require('./pages/home')), // assigned first, replaced by /pages/wrapper
  route('/404', require('./pages/404')),
  route('/privacy', require('./pages/privacy')),
  route('/opensource', require('./pages/opensource')),
])

app.toString('/')  // => renders /pages/wrapper
```

This PR allows the above to be written and processed in the correct order.

```javascript
app.addRoute('/', require('./pages/wrapper'), [ // not assigned
  route('/', require('./pages/home')), // assigned first
  route('/404', require('./pages/404')),
  route('/privacy', require('./pages/privacy')),
  route('/opensource', require('./pages/opensource')),
])

app.toString('/')  // => renders /pages/home under pages/wrapper
```

### DIFF
--- a/docs/images/esm.svg
+++ b/docs/images/esm.svg
@@ -14,7 +14,7 @@
   <g fill="#fff" text-anchor="middle" font-family="DejaVu Sans,Verdana,Geneva,sans-serif" font-size="11">
     <text x="16.5" y="15" fill="#010101" fill-opacity=".3">esm</text>
     <text x="16.5" y="14">esm</text>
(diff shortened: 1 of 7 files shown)


JRJurman:
@chtinahow verbally approves this PR

PR #109, opened 2019-02-17
Fix: populate webEngine on addAction

JRJurman:
## Summary

If you don't `start` the app, then webEngine remains empty, even if `addAction` calls have been made. This means any attempt to call actions or look at store props before the app has started will fail.

```javascript
const app = new Tram({ webEngine: {} })
app.addActions( ... )

console.log(app.webEngine) // -> undefined
```

## The Fix
This PR (among other things) checks if webEngine exists and updates the store and actions on `addActions`. A test has been added to validate this functionality.

## Other Changes
I realized that some old example launchers were still in the repo, so those have been removed. The `xo` dependency has been pinned (since it still remains in pre-release, so `^` was pulling breaking changes). The examples that do exist have been updated to reference the forked dependencies.

### DIFF
--- a/configs/webpack.config.testem.js
+++ b/configs/webpack.config.testem.js
@@ -1,5 +1,6 @@
 module.exports = {
   entry: './tests/browser-runner.js',
+  mode: 'development',
   module: {
     rules: [
       {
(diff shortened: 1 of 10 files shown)


ethanjurman:
:+1:

Issue #117, opened 2019-06-01
Improve Performance by Stopping Render if renderTracker = false


JRJurman:
## Summary

We use a renderTracker to determine if we should morph the dom, but we don't stop from building these components after we've decided on a page to render. We should kill the mount process early by ditching the rendering (probably in dom) and verify that we don't accidently trigger any cleanup effects.

JRJurman:
Closed by #120

PR #120, opened 2019-06-02
new architecture diagram

### LINKED ISSUE #115: Architecture Diagram in github Readme
Since the architecture has changed significantly, it would be valuable to document that in the repo.

JRJurman:
## Summary

Resolves #115 

Included architecture diagram to detail what the different functions are dependent on internally and externally. It's not fully detailed, but should give a general picture of how the pieces are inter-connected.

In the process of building this PR, more integration tests were added, optimizations (#117) were made, tests that verified partial setups (global state without certain values) were removed, and components were renamed (specifically Log and Render Tracker)


JRJurman:
Tatermorph is the external dependency for mount

PR #121, opened 2019-06-20
hot-fix umd build

JRJurman:
## Summary

This PR fixes an issue that caused the umd build to suddenly break.
This PR also introduces integration tests which should validate different flows (these were previously tested in unit tests, but that proved to miss these use-cases). These integration tests validate against the umd build itself.


integration-tests/specs/use-global-spec/index.html
```diff
@@ -0,0 +1,41 @@
...
+  <body>
+    <div class="main"></div>
+    <script>
+      const { registerHtml, useGlobalState, start } = window['tram-one']
+      const rootHtml = registerHtml()
+
+      const login = () => {
+        const [name, setName] = useGlobalState('name', '')
+        if (name === '') {
+          setName('Sprocket')
```
chtinahow: He's going to be immortalized

chtinahow: Merge after circleci fix

PR #122, opened 2019-06-21
Update effect

JRJurman:
## Summary

To better match React's hook's functionality, useEffect now takes in a list of `triggers` which will cause an effect to cleanup and re-trigger if it has been updated.

Luckily, this was a fairly easy and simple change. Included in the PR are integration and unit tests.

### DIFF
--- a/src/hooks/useEffect.js
+++ b/src/hooks/useEffect.js
@@ -1,7 +1,7 @@
 const { TRAM_HOOK_KEY, TRAM_EFFECT_QUEUE } = require('../engine-names')
 const { getEffectStore } = require('../effect-store')
 const { getWorkingKeyValue, incrementWorkingKeyBranch } = require('../working-key')
-const { assertGlobalSpaceAndEngine, assertIsFunction } = require('../asserts')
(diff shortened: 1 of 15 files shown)


src/hooks/useEffect.js
```diff
@@ -40,6 +41,11 @@ module.exports = (globalSpace, storeName = TRAM_EFFECT_QUEUE, workingKeyName = T
     // this makes successive useEffects calls unique (until we reset the key)
     incrementWorkingKeyBranch(globalSpace, workingKeyName)
 
-    effectStore[key] = onEffect
+    // if we have triggers, append them to the key
+    // this will make calls with new / different triggers to restart the effect
+    const formatTriggers = triggers.join(':')
+    const keyWithTriggers = `${key}(${formatTriggers})`
+
+    effectStore[keyWithTriggers] = onEffect
```
JRJurman: this is the meat of the change.
Basically, appending the triggers to the key, which means that if the triggers change, then on finishing the mount, we'll have noticed that one key no longer exists, and a new key that we've never seen before does, causing the respective cleanup and start of the effects.

PR #132, opened 2019-10-26
remove examples and ssl commands

JRJurman:
## Summary

This PR removes the examples in the repo.

I've never been a huge fan of these examples, and I think the information is better captures in the tutorials / codesandbox / api docs. These are just misleading in how you should actually build a Tram-One app. 

Also taking the opportunity to remove some of the ssl commands in the package.json, in reality, these shouldn't have to be run again, and if I need to I can always reference this PR on how to do it.


JRJurman:
If we want to include examples in a different way (with codesandbox or jsfiddle or whatever), I'm super eager to discuss potential options, and we can make an issue to capture that work!

PR #135, opened 2019-12-15
Reactive Hooks and Updates - Major Update - v9 

JRJurman:
## Summary

This PR is a major rewrite / refactoring of how tram-one works, making individual components capable of updating themselves when their state updates. This is a huge performance boost over the existing implementation that mimics React's methodology to re-render the entire app when state changes.

To better understand the motivation behind these changes, watch Rich Harris's talk on Rethinking Reactivity https://youtu.be/AdNJ3fydeao

## Beta Release

Currently, the contents of this branch are released on npm under the beta tag, so you can test these changes by installing `tram-one@beta`. The version should be something like `v9.0.0-beta.0`
```
npm install --save tram-one@beta
```

You can also install directly from this branch:
```
npm install --save Tram-One/tram-one#reactive
```

## Codesandbox
If you want to play around with an example (based on the tutorial), you can see this code sandbox
https://codesandbox.io/s/tram-one-reactive-beta-o6x4j

image

You can even see in the demo above that the components only re-render when their state changes (without useMemo 😄)

### Observables & Reactions
* internal change
* user-facing change

Instead of `useState` and `useGlobalState`, there are two new hooks, `useObservable` and `useGlobalObservable`. These hooks give access to an observable object (created using [nx-js/observer-util](https://github.com/nx-js/observer-util) ) that when updated or set will trigger JUST those components using that state to update. 

This removes the need for `useMemo`, as components will only trigger when the state they are listening to changes. 

Effects are also observed, and so if state that they are dependent on changes, they will independently (from the component) re-trigger as well. This removes the need for dependencies or trigger in the `useEffect`. 

Components and effects are triggered / unobservered using the browser's [MutationObserver](https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver), which tracks which new components are added, removed, or replaced.

### Better Docstrings
* user-facing change

Instead of having adjacent `.doc.js` files, the refactoring (mentioned below) has allowed the code to be properly annotated with jsdocs.

image

### Updated API Docs
* user-facing change

The examples have been updated to include live StackBlitz for each function. This should be more valuable and exhaustive then the existing static examples, and better than the tutorial for giving people an idea for how to use the functions.

image

### Node Properties
* internal change
* _slight_ user-facing change

In order to properly update components individually, we are now storing properties on the mounted DOM. An individual element will have properties like `tram-tag-reaction`, `tram-tag-new-effects`, and `tram-tag-cleanup-effects`. These properties allow us to process the elements in the mutation observer.

### Remove / Revert Dependencies
* internal change

Since we are updating the individual component, we are no longer dependent on `nanomorph` (or it's fork, `tatermorph`).

Since we are using observables, we no longer need to manage state using `hover-engine`.

For some reason we still had `rlite-router` in our dependencies, but that has been extracted out to `use-url-params`. 

### Refactoring Globals
* internal change

Previously all functions were wrapped in a function to take in a global. This has been refactored so that they get whatever the currently set global is from a function. This makes the code easier to read and easier to add to.

```diff
- const registerHtml = globalSpace => registry => {
-	return registerDom(globalSpace)(null, registry)
+ const registerHtml = registry => {
+	return registerDom(null, registry)
```

### Remove Render-lock
* internal change

Because we are no longer doing full sweeping renders of the app, we can remove the render-lock (a mechanism for stopping rendering loops in response to effects).

### Mount mimics component rendering
* internal change
* _slight_ user-facing change

Mount follows similar logic to how we create and process other components. Sadly, for reasons that I mostly understand but can't quite wrap my head around, in order for this to work the app needs to be wrapped in a surrounding element (right now a div). 

### Remove custom asserts / internal asserts
* internal change
* _slight_ user-facing change
Some asserts wouldn't have an impact on the end user, so those have been removed. We are now using the library `type` to manage verifying types.

### testing-library
* project change
instead of unit tests on the functions themselves, we are writing tests that start an app and verify the DOM has the expected results. This is made possible with [Testing Library](https://testing-library.com/), which gives access to the jsdom, and allows us (in unit tests) to query and interact with it like a headless browser.  This is valuable because with code-coverage, it's easier to find dead branches, we are testing the code in a more real environment, and we don't have to mock out internals to verify functions work as expected. 

## Todo
- [x] Investigate using non-forked dependencies (since we no longer need `getEvents`)
- [x] Validate routing updates with url-listener
- [x] Update unit tests
- [x] Update integration tests
- [x] Update Website Docs
- [x] Update Tutorial
- [x] Resolve Build Warnings

## Post-Release Todo
- [ ] Update tram-one-express


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "tram-one",
-  "version": "8.1.3",
+  "version": "9.0.0-beta.2",
```
ethanjurman: is it okay to have the beta version in a master branch? or is the versioning managed by npm?

src/hooks/observableHook.js
```diff
@@ -0,0 +1,43 @@
...
+	}
+
+	// generate getter for key
+	const keyGetter = observableStore[resolvedKey]
+
+	// generate setter for key
+	const isObject = typeof value === 'object'
+	const keySetter = newValue => {
+		if (isObject) console.warn('Tram-One: Avoid using setters if dealing with objects or arrays')
+		observableStore[resolvedKey] = newValue
```
ethanjurman: I think this `if` block is highly confusing- but if it's common syntax for you :+1:

src/tram-space/index.js
```diff
@@ -0,0 +1 @@
+module.exports = require('./tram-space')
```
ethanjurman: this kinda syntax is not contributing to a folder/ file clutter is it?
Recently I was in the camp of "there is no such thing as too many folders" and now I feel like i'm double backing a lot on that, well names files (that organize themselves alphabetically) and low folder count makes it a lot easier to manage and take in.

package.json
JRJurman: When we do the final merge, it'll be just 9.0.0

src/hooks/observableHook.js
JRJurman: I might add more here, in which case I will make it a block... But yeah, I use one line ifs in a couple of places

src/tram-space/index.js
JRJurman: My biggest goal is to keep test and source files organized. The `index.js` just makes it so imports are easier.

I'm not adverse to a different pattern if you have an example / suggestion.

I agree that some of the folders make it... Difficult.

JRJurman: Since test files are (and will continue to be) in a separate folder, I may remove the folders and have a flatter structure.

JRJurman:
I can't publish the site on circle ci (or rather, the page doesn't render the way it should), so this is what it looks like
image

src/mount.js
```diff
@@ -0,0 +1,42 @@
...
+ * Updates a selector with an initial component for the first render.
+ */
+module.exports = (selector, component) => {
+	ensureValue(selector, { errorMessage: `Tram-One: selector should be defined as a CSS Selector or DOM Node, recieved ${selector}` })
+	ensureFunction(component, { errorMessage: `Tram-One: component should be a function, recieved ${typeof selector}, ${selector}` })
+
+	// if the selector is a string, try to find the element,
+	// otherwise it's probably DOM that we should write directly to
+	const target = (typeof selector) === 'string' ? document.querySelector(selector) : selector
+	if (target === null) {
```
chtinahow: do you have to be explicit rather than !target?

src/mutation-observer.js
```diff
@@ -0,0 +1,87 @@
...
+const processEffects = node => {
+	const hasNewEffects = node[TRAM_TAG_NEW_EFFECTS]
+
+	if (hasNewEffects) {
+		// create an array for the cleanup effects
+		node[TRAM_TAG_CLEANUP_EFFECTS] = []
+
+		// run all the effects, saving any cleanup functions to the node
+		node[TRAM_TAG_NEW_EFFECTS].forEach(effect => {
+			let cleanup
```
chtinahow: ew

configs/rollup.config.cjs.js
```diff
@@ -10,7 +10,7 @@ const plugins = [
 ]
 
 export default {
-	input: 'src/tram-one/index.js',
+	input: 'src/tram-one',
```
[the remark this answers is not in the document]
JRJurman: we're no longer using `index.js` because we no longer have tests in the folders, so most folders were just the source and an index file.  It made functions unnecessarily nested, and it's cleaner now without them (err, at least flatter)

integration-tests/mock-components/graphic.js
```diff
@@ -0,0 +1,8 @@
+const { registerSvg } = require('../../src/tram-one')
+const svg = registerSvg()
+
+module.exports = () => svg`
+<svg viewBox="0 0 864 864">
+	<circle data-testid="graphic-circle" fill="#FDF491" cx="100" cy="100" r="20"/>
+</svg>
+`
```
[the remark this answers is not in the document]
JRJurman: good catch. ideally this stuff would be caught by linters, but alas, that appears to be... difficult... see https://github.com/Tram-One/eslint-config-tram-one/issues/7

integration-tests/mock-components/home.js
```diff
@@ -0,0 +1,9 @@
+const { registerHtml } = require('../../src/tram-one')
+const html = registerHtml()
+
+module.exports = (props, children) => html`
+	<div data-testid="home-page">
```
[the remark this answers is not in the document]
JRJurman: this is actually one `tab`, which can be configured in the github ui with query params (I don't know what they are off-hand, but they exist).

integration-tests/mock-components/updated-effect.js
```diff
@@ -0,0 +1,13 @@
...
+
+module.exports = ({ initialCount }) => {
+	const [counter] = useObservable({ count: Number(initialCount) })
+	const increment = () => { counter.count += 1 }
+	useEffect(() => {
+		document.title = `The count is ${counter.count}`
+		return () => { document.title = `The count was ${counter.count}` }
+	})
+
+	return html`<button data-testid="updatable-button" onclick=${increment}>${counter.count}</button>`
```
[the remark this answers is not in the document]
JRJurman: a large chunk of this is just in the attrs, so I don't know if it would be smaller to wrap this. More often I do that if the child is another element. In this case, it's just text.

src/mount.js
JRJurman: `document.querySelector` specifically returns `null` when an element can't be found... This doesn't do all the type checking I would like, but casting that value to a boolean isn't really required.

src/mutation-observer.js
JRJurman: this is actually an appropriate (if a bit strange), use of `let`, where it can be read before it's assigned. There are a few `let`s after the refactor, and they have to do with how the observer gets called (and it's scope).

PR #141, opened 2020-07-24
Fixes and Improvements for Hooks

JRJurman:
## Summary
This PR contains a set of changes that were implemented when working on the beer-cart.io project. There were later validated using tests to confirm that they were in fact required (and useful).

## Changes

update mutation-observer.js
* copy validation done on effect cleanup function for effect updates

update observable-hook.js
* add third index in the result array, for a non-proxied value that doesn't trigger updates
* check that the value being written is defined before writing a value to the store

update tram-one.js
* expose raw.js as top-level function

udpate integration.test.js
* remove unused import
* add new test for raw method

create regression.test.js
* add regression test to verify cleanups 
* add regression test to verify that global obversables are not overwritten with undefined

Misc changes to support the above:
* create conflicting-global-counter.js
* create frozen-counter.js
* update mock-components/index.js
* create updatable-promise-page.js
* create updated-promise.js

### DIFF
--- a/integration-tests/integration.test.js
+++ b/integration-tests/integration.test.js
@@ -1,10 +1,10 @@
-const { getByTestId, fireEvent, wait } = require('@testing-library/dom')
+const { getByTestId, getByRole, fireEvent, wait } = require('@testing-library/dom')
 const { start, registerHtml } = require('../src/tram-one')
 
-const { globalCounter, graphic, home, loadingPage, loading, removablePage, removable, updatablePage, updatable, updatedEffect, urlLabel } = require('./mock-components')
+const { frozenCounter, globalCounter, graphic, home, loadingPage, loading, removablePage, updatablePage, updatable, urlLabel } = require('./mock-components')
 
 const html = registerHtml({
-	globalCounter, graphic, home, loadingPage, loading, removablePage, removable, updatablePage, updatable, updatedEffect, urlLabel
+	globalCounter, home, loadingPage, loading, updatable, urlLabel
 })
 
 describe('Tram-One', () => {
@@ -281,5 +281,19 @@ describe('Tram-One', () => {
 		// verify that the global tramSpace was used
 		const keys = Object.keys(global.tramSpace)
 		expect(keys.length).toBeGreaterThan(0)
+	})
+
+	it('should not trigger side-effects of raw objects', () => {
+		// mount the app on the container
+		const container = document.createElement('div')
+		start(container, frozenCounter)
+
+		// verify that the button is rendered with the default text
+		expect(getByRole(container, 'button')).toHaveTextContent('clicks: 0')
+		// click on the click button to trigger update
+		fireEvent.click(getByRole(container, 'button'))
+
+		// verify that the button was not updated
+		expect(getByRole(container, 'button')).toHaveTextContent('clicks: 0')
 	})
 })
--- a/integration-tests/integration.test.js
+++ b/integration-tests/integration.test.js
@@ -1,4 +1,4 @@
-const { getByTestId, getByRole, fireEvent, wait } = require('@testing-library/dom')
+const { getByTestId, getByRole, fireEvent, waitFor } = require('@testing-library/dom')
 const { start, registerHtml } = require('../src/tram-one')
 
 const { frozenCounter, globalCounter, graphic, home, loadingPage, loading, removablePage, updatablePage, updatable, urlLabel } = require('./mock-components')
@@ -92,7 +92,7 @@ describe('Tram-One', () => {
 		expect(getByTestId(container, 'home-page')).toHaveTextContent('loading')
 
 		// wait for mutation observer to pick up new elements
-		await wait()
+		await waitFor(() => {})
 
 		// verify that it (eventually) has the finished text
 		expect(getByTestId(container, 'home-page')).toHaveTextContent('Finished')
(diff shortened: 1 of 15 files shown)


integration-tests/integration.test.js
```diff
@@ -282,4 +282,18 @@ describe('Tram-One', () => {
...
+		start(container, frozenCounter)
+
+		// verify that the button is rendered with the default text
+		expect(getByRole(container, 'button')).toHaveTextContent('clicks: 0')
+		// click on the click button to trigger update
+		fireEvent.click(getByRole(container, 'button'))
+
+		// verify that the button was not updated
+		expect(getByRole(container, 'button')).toHaveTextContent('clicks: 0')
+	})
```
JRJurman, line 298: This test verifies that a raw (non-proxied) object will not cause side effects when it is mutated.

integration-tests/mock-components/conflicting-global-counter.js
```diff
@@ -0,0 +1,23 @@
...
+const html = registerHtml()
+
+module.exports = () => {
+	// example where it is undefined first
+	useGlobalObservable('global-counter', undefined)
+	const [counter] = useGlobalObservable('global-counter', 5)
+
+	// example where it is defined first
+	useGlobalObservable('global-counter-2', 5)
+	const [counter2] = useGlobalObservable('global-counter-2', undefined)
```
JRJurman, line 11: these lines verify that regardless of the order, a global-observable will not be overridden with `undefined`. This is more common when we want to depend on one component setting a value, and somewhere else in the code we want to just read it (and we have no idea what the default value might be).

```javascript
// somewhere.js
const [setupValue] = useGlobalObservable('global-value', 5)
```

```javascript
// somewhere-else.js
const [readValue] = useGlobalObservable('global-value')
```

integration-tests/mock-components/frozen-counter.js
```diff
@@ -0,0 +1,20 @@
...
+const html = registerHtml()
+
+module.exports = () => {
+	const [,, rawCounter] = useObservable({ now: { clicks: 0 } })
+	const onclick = () => {
+		// while a spread operator would work for single nested objects
+		// for objects that are deeply nested, we need to use raw to
+		// transform the entire object to it's raw value
+		rawCounter.now.clicks += 1
+	}
```
JRJurman, line 11: Basically what this comment says... In reality, this may be rarely used, but for beer-cart.io, it was required to pass and interact with an object, and then wholesale commit to changing it or not.

It appears the that having observables change the UI on mutation can be a double edged sword.

integration-tests/mock-components/updatable-promise-page.js
```diff
@@ -0,0 +1,27 @@
...
+
+	const removeCount = () => {
+		setShowCount(false)
+	}
+
+	return html`
+		<home>
+			<button data-testid="increment-count" onclick=${increment}>Increment</button>
+			<button data-testid="remove-count" onclick=${removeCount}>Remove Count</button>
+			${showCount ? html`<updated-effect initialCount=${count} />` : 'Counter Removed'}
```
JRJurman, line 24: the whole purpose of this component is to be able to trigger the cleanup of a component... in reality, this should be refactored so that all kinds of cleanup effects can be validated... something that should happen when the test code is cleaned up (in another PR... probably).

integration-tests/mock-components/updated-promise.js
```diff
@@ -0,0 +1,12 @@
+const { registerHtml, useObservable, useEffect } = require('../../src/tram-one')
+const html = registerHtml()
+
+module.exports = ({ initialCount }) => {
+	const [counter] = useObservable({ count: Number(initialCount) })
+	const increment = () => { counter.count += 1 }
+	useEffect(async () => {
+		document.title = `The count is ${counter.count}`
+	})
```
JRJurman, line 9: yes! `useEffect` supports async... well, it should have, now it does for sure-sure.

integration-tests/regression.test.js
```diff
@@ -0,0 +1,39 @@
+const { getByTestId, fireEvent, wait } = require('@testing-library/dom')
+const { start } = require('../src/tram-one')
+
+const { updatablePromisePage, conflictingGlobalCounter } = require('./mock-components')
+
+describe('Tram-One - regressions', () => {
```
JRJurman, line 6: this file represents things that aren't necessarily features, but they are requirements around how tram-one should act... really it comes out of tram-one acting in strange ways that are bugs, and we want to make sure they don't happen again -> ergo regressions

integration-tests/regression.test.js
```diff
@@ -0,0 +1,39 @@
...
+		// verify the title is the default value
+		expect(document.title).toBe('The count is 5')
+
+		// previously this would fail because the cleanup was called,
+		// even though it was not a function, and instead was a promise (the result of an async function)
+		// the fix was to do the same check we do in other places, to make sure cleanup is a function
+		expect(() => {
+			// click on the inner component (that updates the component's state)
+			fireEvent.click(getByTestId(container, 'updatable-button'))
+		}).not.toThrow()
```
JRJurman, line 27: basically this comment... the `expect().not.toThrow()` is neat!

package.json
```diff
@@ -86,7 +86,8 @@
       "valid-jsdoc": 0,
       "capitalized-comments": 0,
       "new-cap": 0,
-      "unicorn/no-process-exit": 0
+      "unicorn/no-process-exit": 0,
+      "unicorn/no-unreadable-array-destructuring": 0
```
JRJurman, line 90: now that 
```
[,, nonProxiedValue] = useObservable({ value: 10 })
```
is possible, we want to disable this rule... eventually this should be rolled up into the eslint-config - https://github.com/Tram-One/eslint-config-tram-one/issues/12

PR #151, opened 2021-02-03
v10: major refactoring

JRJurman:
## Summary
It's that time of year again where Tram-One gets another major (breaking) refactoring :grin: 

Commits are split up to make reviewing easier, here are the overall descriptions for each commit:

## v10: linting & jsdoc changes
xo updated, and some of the other changes in the PR caused linting to happen. This commit can be safely ignored.

## v10: remove global tramSpace, split up logic for building mounting container
Removing `global.tramSpace` as a way for people to change where tram-one mounted. This was a strange interface, and locking down to "the window is where data is stored" is probably safer and easier to understand going forward. There was an attempt to get the data to be stored on the top-level element, but because of how components are processed, this was not possible.

## v10: remove useObservable, add useStore
Probably the most significant change in the PR, this change removes the ability to use primitive data in the state hook (e.g. booleans, integers, strings). The reason for this is that `useEffect` wasn't able to understand when a dependent hook was updated if it was a primitive value. Also, the "two different ways to update state" was probably more confusing than it should have been. To make everything more consistent, the state hook in tram-one now **only takes in objects**.  To make this breaking change more obvious (and because the previous hook name was a little long anyways), the new hook is named `useStore`.

Also part of this change, updated the interface for useUrlParams. Instead of returning false, it always returns an object, which has a key `{ matches }`

## v10: new integration tests
Updated the integration and regression tests to test against a single application. This made it clearer what was being tested, without rebuilding an app every time. This reduced the total number of test components as well.

## bump versions, update badges
Bump to version 10, and update the build badges 

## Checklist
- [x] PR Summary
- [x] Tests
- [x] Version Bump


docs/badges/cjs.svg
```diff
@@ -14,7 +14,7 @@
   <g fill="#fff" text-anchor="middle" font-family="DejaVu Sans,Verdana,Geneva,sans-serif" font-size="11">
     <text x="16.5" y="15" fill="#010101" fill-opacity=".3">cjs</text>
     <text x="16.5" y="14">cjs</text>
-    <text x="59" y="15" fill="#010101" fill-opacity=".3">27.28 kB</text>
-    <text x="59" y="14">27.28 kB</text>
+    <text x="59" y="15" fill="#010101" fill-opacity=".3">25.58 kB</text>
+    <text x="59" y="14">25.58 kB</text>
```
JRJurman, line 18: :smile:

integration-tests/integration.test.js
```diff
@@ -1,299 +1,151 @@
-const { getByTestId, getByRole, fireEvent, waitFor } = require('@testing-library/dom')
-const { start, registerHtml } = require('../src/tram-one')
-
-const { frozenCounter, globalCounter, graphic, home, loadingPage, loading, removablePage, updatablePage, updatable, urlLabel } = require('./mock-components')
-
-const html = registerHtml({
-	globalCounter, home, loadingPage, loading, updatable, urlLabel
-})
+const { getByText, fireEvent, waitFor } = require('@testing-library/dom')
+const { startApp } = require('./test-app')
```
JRJurman, line 2: Rather than importing and building a bunch of isolated components, we have a mock app that exercises tram-one in a variety of ways.

JRJurman: As a result of this refactoring, all the mock-components made for testing have been removed, favoring instead a single app with a fewer number of components.

src/tram-space.js
```diff
@@ -1,21 +1,22 @@
...
+		window['tram-space'] = window['tram-space'] || {}
+		return window['tram-space']
+	}
 
 	// if tramSpace and window are undefined,
 	// set the globalSpace to a new empty object
-	global.tramSpace = {}
-	return global.tramSpace
+	global['tram-space'] = {}
+	return global['tram-space']
```
JRJurman, line 21: Should remove references to `global['tram-space']`

ethanjurman (approved):
LGTM

chtinahow (approved):
Looks good 👍

PR #152, opened 2021-02-08
Fix Array in useStore (v10.0.1)

JRJurman:
## Summary

In the release of v10, the ability to store arrays was broken (because to make a shallow copy, the spread syntax for objects was used, but this inadvertently changed arrays to objects). There was no regression test for this, so a new one has been added. 

## Checklist
- [x] PR Summary
- [x] Tests
- [x] Version Bump

### DIFF
--- a/src/use-global-store.js
+++ b/src/use-global-store.js
@@ -10,7 +10,7 @@ const observableHook = require('./observable-hook')
  * it will cause only the components that are dependent on that value to update.
  *
  * @param {string} key a unique string to write and read the global value
- * @param {Object|Array} value the default value to start the state at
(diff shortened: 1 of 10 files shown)


src/use-global-store.js
```diff
@@ -10,7 +10,7 @@ const observableHook = require('./observable-hook')
  * it will cause only the components that are dependent on that value to update.
  *
  * @param {string} key a unique string to write and read the global value
- * @param {Object|Array} value the default value to start the state at
+ * @param {Object|Array} defaultValue the default value to start the store at
```
JRJurman, line 13: also taking this opportunity to change the parameter jsdoc here, it should be default value, as that is what it is.

Issue #161, opened 2021-07-04
Inputs lose focus when sharing state


JRJurman:
## Summary

In the following repo, there are issues with keeping input. This might have to do with shared global state :thinking: 
https://github.com/JRJurman/tram-one-input-testing

## Test
Run `npm start`
Start typing

### Expected
Browser does not lose focus while typing

### Actual
On every key press, input loses focus

### Notes
Not sure if this input tag is being wired up correctly, or if doing onkeyup is discouraged, but I feel like that _should_ be supported.

JRJurman:
Verified fixed in 10.1.15 :+1:

PR #163, opened 2021-07-12
Rewrite of Focus Logic

JRJurman:
## Summary
Moved the focus logic to happen when the element is being recreated, rather than at the mutation observer level. This resolves the issues we were seeing in #161 and is cleaner than the previous implementation (which required writing data to elements and tracking focus ourselves).

## Checklist
- [x] PR Summary
- [x] Tests
- [x] Version Bump


docs/badges/umd.svg
```diff
@@ -14,7 +14,7 @@
   <g fill="#fff" text-anchor="middle" font-family="DejaVu Sans,Verdana,Geneva,sans-serif" font-size="11">
     <text x="16.5" y="15" fill="#010101" fill-opacity=".3">umd</text>
     <text x="16.5" y="14">umd</text>
-    <text x="59" y="15" fill="#010101" fill-opacity=".3">21.16 kB</text>
-    <text x="59" y="14">21.16 kB</text>
+    <text x="59" y="15" fill="#010101" fill-opacity=".3">21.02 kB</text>
+    <text x="59" y="14">21.02 kB</text>
```
JRJurman, line 18: We actually saved space with the new implementation :tada:

chtinahow (approved):
LGTM, How many days did it take you to solve the issue 😆

src/mutation-observer.js
```diff
@@ -63,62 +63,26 @@ const clearNode = node => {
...
-	mutationList
+	const newNodes = mutationList
 		.flatMap(mutation => [...mutation.addedNodes])
-		.flatMap(node => [...(node.querySelectorAll ? node.querySelectorAll('*') : [])])
-		.forEach(processEffects)
+		.flatMap(childrenNodes)
 
-	// if any node had focus, get that element and re-add focus to the new node
-	reapplyFocus(mutationList)
+	newNodes.forEach(processEffects)
```
JRJurman, line 85: this logic didn't actually change, we just made it more clear by making a new `childrenNodes()` function (since that's all this logic interacts with).

JRJurman: To be honest, I'm not actually sure why this only works with just children nodes, or if that's actually intentional. I feel like this might bite us in the butt later, but it doesn't work with the node and children, and this does work, so eh :man_shrugging:

src/observe-tag.js
```diff
@@ -13,9 +13,28 @@ module.exports = tagFunction => {
...
+			removedElementWithFocusData.index = parentAndChildrenNodes.findIndex(element => element === document.activeElement)
+
+			// if an element had focus, copy over all the selection data (so we can copy it back later)
+			if (removedElementWithFocusData.index >= 0) {
+				const removedElementWithFocus = parentAndChildrenNodes[removedElementWithFocusData.index]
+				removedElementWithFocusData.selectionStart = removedElementWithFocus.selectionStart
+				removedElementWithFocusData.selectionEnd = removedElementWithFocus.selectionEnd
+				removedElementWithFocusData.selectionDirection = removedElementWithFocus.selectionDirection
+			}
+
```
JRJurman, line 37: this is the actual logic that does the fix (and the block below).

integration-tests/regression.test.js
```diff
@@ -163,4 +163,59 @@ describe('Tram-One - regressions', () => {
...
+		// verify that the element has focus (before we start changing text)
+		await waitFor(() => {
+			expect(getByLabelText(appContainer, 'Sub Mirror Input')).toHaveFocus()
+		})
+
+		// update the state by typing
+		userEvent.type(getByLabelText(appContainer, 'Sub Mirror Input'), ' Again')
+
+		// verify the element has the new value
+		expect(getByLabelText(appContainer, 'Sub Mirror Input')).toHaveValue('Test Again')
```
ethanjurman, line 211: I can't help but feel like this test gets into a bit of the internals of the side effects, and less about what is critical for the test. I think it's fine, but years down the line will this part of the test make sense?

src/observe-tag.js
```diff
@@ -13,9 +13,28 @@ module.exports = tagFunction => {
...
+			index: null,
+			selectionStart: null,
+			selectionEnd: null,
+			selectionDirection: null
+		}
 
 		// remove oldTag first so that we unobserve before we re-observe
 		if (oldTag) {
+			// if there was focus, we need to figure out what element has it
+			const children = oldTag.querySelectorAll('*')
```
ethanjurman, line 26: Is focus the only data that needs this kind of special casing? Should we look out for other similar meta data that needs to be kept between updates? (Idk if undo states or anything else is special in this regard)

src/observe-tag.js
```diff
@@ -35,6 +54,24 @@ module.exports = tagFunction => {
 		if (oldTag) {
 			oldTag.replaceWith(tagResult)
 
+			// if an element had focus, reapply it
+			if (removedElementWithFocusData.index >= 0) {
+				const children = tagResult.querySelectorAll('*')
+
+				const elementToGiveFocus = [tagResult, ...children][removedElementWithFocusData.index]
```
ethanjurman, line 61: It's creating this array expensive? Should we do a check for 0 and return the tagResult otherwise index +1 on the children array? Or is this operation not frequently called/ not expensive?

ethanjurman (approved):
Looks good! A few questions but glad this overall improves the codebase.

integration-tests/regression.test.js
JRJurman: `toHaveValue` is totally fine, that is expected for an input to work.
Nothing here is working around the internals, it's all just labels and inputs (as you might interact with them in react)

integration-tests/test-app/mirror-input.js
```diff
@@ -0,0 +1,24 @@
+const { registerHtml, useGlobalStore } = require('../../src/tram-one')
```
ethanjurman, line 1: This is more of a personal nit, but I generally try to avoid having these reusable separated test files, as it becomes frustrating when trying to work on additional use cases. You have to check if the test file is used anywhere else, or if it's okay to use our modify it for other tests... I just find it simpler to include it with the test file. But I don't know how other tests here are set up and if that would be an annoying amount of additional code.

src/observe-tag.js
JRJurman: I don't know of other ones off-hand... if there are we can make issues for them. 
Focus might just be the most obvious. If another shows up, then we can look at making a more generic solution.

integration-tests/regression.test.js
ethanjurman: I meant more so that all this test should be checking for is that focus is retained. By having more code it could make it annoying to see where it failed at a glance. But I see what you're saying.

src/observe-tag.js
JRJurman: :thinking: potentially... it could get expensive if the element being re-rendered has like, a ton of children. This part luckily only gets called if the element (or one of it's children) has focus. 

I need to do more generic performance testing, and this could be one of the things that crops up. Without having that performance testing now, it's hard to say for sure.

integration-tests/test-app/mirror-input.js
ethanjurman: Oof nvm. If these are integration tests then this is fine. Thought this was unit tests

JRJurman: previously I had an app for each test, and that was way more of a nightmare to manage (potentially because I was working on breaking changes, but also because it wasn't clear or obvious which app did what).

Ideally, I'd have a single app that fully tested all of tram-one, that was semantically obvious and meaningful. That's obviously hard to come up with...

integration-tests/regression.test.js
JRJurman: oh, yeah, I guess, I was seeing two issues, but I can update the test to also verify that the focus stays (I do that in the test before as well, so I have it already setup).

integration-tests/test-app/mirror-input.js
JRJurman: I'm using a unit test framework to basically write integration tests

Issue #156, opened 2021-04-25
Global State should not be overridden in multiple calls


JRJurman:
## Summary

When calling `useGlobalStore` in multiple components with the same key, we should make sure that the values are either merged or sanely not overridden (potentially we should have a runtime error if we see the key being written to by multiple components?)

PR #164, opened 2021-07-18
Fix input focus on re-ordered elements

JRJurman:
## Summary
Fixes an issue where if the elements were re-ordered or changed significantly, we wouldn't be able to re-attach focus. Worse so, we never check if the element exists before trying to focus it, so we would get an error!

This PR also fixes an error where the scroll position would be reset when typing in an input. 

## Screenshots (Before)
When interacting with an input that has `<span>-</span>` for every letter in the input. This is what happens when typing and then deleting (this is the integration test app, which can be run locally with `npm run test:app`):
image
The first set of errors is because we are trying to `setSelectionRange` on a `<span>`, the second error is because after deleting no element exists at the index that the `<input>` was on.

## Screenshots (After)
No errors :tada: 
image

## Checklist
- [x] PR Summary
- [x] Tests
- [x] Version Bump

### DIFF
--- a/docs/badges/cjs.svg
+++ b/docs/badges/cjs.svg
@@ -14,7 +14,7 @@
   <g fill="#fff" text-anchor="middle" font-family="DejaVu Sans,Verdana,Geneva,sans-serif" font-size="11">
     <text x="16.5" y="15" fill="#010101" fill-opacity=".3">cjs</text>
     <text x="16.5" y="14">cjs</text>
-    <text x="59" y="15" fill="#010101" fill-opacity=".3">27.59 kB</text>
-    <text x="59" y="14">27.59 kB</text>
+    <text x="59" y="15" fill="#010101" fill-opacity=".3">29.14 kB</text>
+    <text x="59" y="14">29.14 kB</text>
   </g>
 </svg>
--- a/docs/badges/cjs.svg
+++ b/docs/badges/cjs.svg
@@ -14,7 +14,7 @@
   <g fill="#fff" text-anchor="middle" font-family="DejaVu Sans,Verdana,Geneva,sans-serif" font-size="11">
     <text x="16.5" y="15" fill="#010101" fill-opacity=".3">cjs</text>
     <text x="16.5" y="14">cjs</text>
-    <text x="59" y="15" fill="#010101" fill-opacity=".3">28.96 kB</text>
-    <text x="59" y="14">28.96 kB</text>
+    <text x="59" y="15" fill="#010101" fill-opacity=".3">29.15 kB</text>
+    <text x="59" y="14">29.15 kB</text>
   </g>
 </svg>
--- a/docs/badges/cjs.svg
+++ b/docs/badges/cjs.svg
@@ -14,7 +14,7 @@
   <g fill="#fff" text-anchor="middle" font-family="DejaVu Sans,Verdana,Geneva,sans-serif" font-size="11">
     <text x="16.5" y="15" fill="#010101" fill-opacity=".3">cjs</text>
     <text x="16.5" y="14">cjs</text>
-    <text x="59" y="15" fill="#010101" fill-opacity=".3">29.15 kB</text>
-    <text x="59" y="14">29.15 kB</text>
+    <text x="59" y="15" fill="#010101" fill-opacity=".3">29.33 kB</text>
+    <text x="59" y="14">29.33 kB</text>
   </g>
 </svg>
--- a/docs/badges/cjs.svg
+++ b/docs/badges/cjs.svg
@@ -14,7 +14,7 @@
   <g fill="#fff" text-anchor="middle" font-family="DejaVu Sans,Verdana,Geneva,sans-serif" font-size="11">
     <text x="16.5" y="15" fill="#010101" fill-opacity=".3">cjs</text>
     <text x="16.5" y="14">cjs</text>
-    <text x="59" y="15" fill="#010101" fill-opacity=".3">29.14 kB</text>
-    <text x="59" y="14">29.14 kB</text>
+    <text x="59" y="15" fill="#010101" fill-opacity=".3">28.96 kB</text>
+    <text x="59" y="14">28.96 kB</text>
   </g>
 </svg>
--- a/integration-tests/regression.test.js
+++ b/integration-tests/regression.test.js
@@ -1,4 +1,4 @@
-const { getByText, fireEvent, waitFor, getByLabelText } = require('@testing-library/dom')
+const { getByText, queryAllByText, fireEvent, waitFor, getByLabelText } = require('@testing-library/dom')
 const { default: userEvent } = require('@testing-library/user-event')
 const { startApp } = require('./test-app')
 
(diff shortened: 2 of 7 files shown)


src/observe-tag.js
```diff
@@ -1,6 +1,15 @@
...
 const { TRAM_TAG_REACTION, TRAM_TAG_NEW_EFFECTS, TRAM_TAG_CLEANUP_EFFECTS } = require('./node-names')
 
+// helper functions for sorting
+const toIndicies = (node, index) => index
+const toNodes = elementAndChildren => index => elementAndChildren[index]
+const byDistanceFromIndex = targetIndex => (indexA, indexB) => {
+	const diffFromTargetA = Math.abs(indexA - targetIndex)
+	const diffFromTargetB = Math.abs(indexB - targetIndex)
+	return diffFromTargetA < diffFromTargetB ? -1 : 1
+}
```
JRJurman, line 11: this sorting function prioritizes indices that are closest to a target.
E.g. if the target is `5` out of `[1, 2, 3, 4, 5, 6, 7, 8, 9]`, 
then the new sorting would be `[5, 4, 6, 3, 7, 2, 8, 1, 9]`. 

I don't know if there's a more elegant way to write this, but this works :+1:

src/observe-tag.js
```diff
@@ -30,6 +40,7 @@ module.exports = tagFunction => {
 			// if an element had focus, copy over all the selection data (so we can copy it back later)
 			if (removedElementWithFocusData.index >= 0) {
 				const removedElementWithFocus = parentAndChildrenNodes[removedElementWithFocusData.index]
+				removedElementWithFocusData.tagName = removedElementWithFocus.tagName
```
JRJurman, line 43: since we now search for the element we previously had focus on, we want to store the `tagName` (this way, if an input is surrounded by other different elements, we can still find it again and reapply focus)

docs/badges/cjs.svg
```diff
@@ -14,7 +14,7 @@
   <g fill="#fff" text-anchor="middle" font-family="DejaVu Sans,Verdana,Geneva,sans-serif" font-size="11">
     <text x="16.5" y="15" fill="#010101" fill-opacity=".3">cjs</text>
     <text x="16.5" y="14">cjs</text>
-    <text x="59" y="15" fill="#010101" fill-opacity=".3">27.59 kB</text>
-    <text x="59" y="14">27.59 kB</text>
+    <text x="59" y="15" fill="#010101" fill-opacity=".3">28.96 kB</text>
+    <text x="59" y="14">28.96 kB</text>
```
JRJurman, line 18: most of this is additional comments, which is good :+1:

integration-tests/regression.test.js
```diff
@@ -221,4 +221,44 @@ describe('Tram-One - regressions', () => {
 		// cleanup - remove app
 		appContainer.remove()
 	})
+
+	it('should not error when reseting focus if the number of elements changed', async () => {
```
JRJurman, line 225: the test app needed to be updated, and actually caused a previous test to fail without the fix, but it still feels worthwhile to give this it's own test, so that if they ever get refactored, we don't forget about it.

integration-tests/test-app/sub-mirror-input.js
```diff
@@ -8,14 +8,17 @@ const html = registerHtml({
  * component to test url parameters
  */
 module.exports = () => {
-	const mirrorable = useGlobalStore('mirrorable-input', { value: '' })
```
JRJurman, line 11: I don't think we actually needed this `, { value: '' }`, it was documented as an error #156 , but this works without it

JRJurman: In reality, I just need to create a minimal example of this...

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "tram-one",
-  "version": "10.1.5",
+  "version": "10.1.6",
```
JRJurman, line 3: no API change, fixes an issue with a feature implementation, so patch is good :+1:

src/observe-tag.js
```diff
@@ -15,24 +33,32 @@ module.exports = tagFunction => {
...
+				// get the actual element
+				const removedElementWithFocus = allElements[removedElementWithFocusData.index]
+
+				// copy over the data
+				removedElementWithFocusData.tagName = removedElementWithFocus.tagName
 				removedElementWithFocusData.selectionStart = removedElementWithFocus.selectionStart
 				removedElementWithFocusData.selectionEnd = removedElementWithFocus.selectionEnd
 				removedElementWithFocusData.selectionDirection = removedElementWithFocus.selectionDirection
+				removedElementWithFocusData.scrollLeft = removedElementWithFocus.scrollLeft
+				removedElementWithFocusData.scrollTop = removedElementWithFocus.scrollTop
```
JRJurman, line 61: this preserves the scroll position in an input or textarea that you are typing in. Previously you this would get reset on re-render, so this experience is more in-line with what people would expect. 

I don't know that there's a great way to test this, so I'm fine not writing an integration test - this is one thing where I think it really requires manual verification... anything else would be just checking properties, which seems disingenuous since all elements support `scrollLeft` and `scrollTop`, even if they don't do anything.

src/observe-tag.js
```diff
@@ -56,19 +82,35 @@ module.exports = tagFunction => {
...
+				// if it did, we'll look at the elements closest to the original position
+				const nodeMatchesTagName = node => node.tagName === removedElementWithFocusData.tagName
+				const elementToGiveFocus = allElements
+					.map(toIndicies)
+					.sort(byDistanceFromIndex(removedElementWithFocusData.index))
+					.map(toNodes(allElements))
+					.find(nodeMatchesTagName)
+
+				// if the element / child exists, focus it
+				if (elementToGiveFocus !== undefined) {
```
JRJurman, line 98: This is the actual fix. We look through all elements now, and verify that the element even exists. Previously, even if the element was removed, we'd still try to focus on something, which caused errors to be thrown.

docs/badges/cjs.svg
```diff
@@ -14,7 +14,7 @@
   <g fill="#fff" text-anchor="middle" font-family="DejaVu Sans,Verdana,Geneva,sans-serif" font-size="11">
     <text x="16.5" y="15" fill="#010101" fill-opacity=".3">cjs</text>
     <text x="16.5" y="14">cjs</text>
-    <text x="59" y="15" fill="#010101" fill-opacity=".3">27.59 kB</text>
-    <text x="59" y="14">27.59 kB</text>
+    <text x="59" y="15" fill="#010101" fill-opacity=".3">29.33 kB</text>
+    <text x="59" y="14">29.33 kB</text>
```
JRJurman, line 18: Most of this is comments, which is fine :+1:

src/observe-tag.js
```diff
@@ -1,6 +1,24 @@
...
+// functions to go to nodes or indicies (made for .map)
+const toIndicies = (node, index) => index
+const toNodes = allNodes => index => allNodes[index]
+
+// sorting function that prioritizes indicies that are closest to a target
+// e.g. target = 3, [1, 2, 3, 4, 5] => [3, 2, 4, 1, 5]
+const byDistanceFromIndex = targetIndex => (indexA, indexB) => {
+	const diffFromTargetA = Math.abs(indexA - targetIndex)
+	const diffFromTargetB = Math.abs(indexB - targetIndex)
+	return diffFromTargetA < diffFromTargetB ? -1 : 1
```
ethanjurman, line 13: Can't you just do `return diffFromTargetA - diffFromTargetB` at this line?

src/observe-tag.js
```diff
@@ -1,6 +1,24 @@
...
+}
+
+// get an array including the element and all it's children
+const parentAndChildrenElements = node => {
+	const children = node.querySelectorAll('*')
+	return [node, ...children]
+}
```
ethanjurman, line 20: Is this recursive/ should it be?

src/observe-tag.js
```diff
@@ -15,24 +33,32 @@ module.exports = tagFunction => {
...
 		}
 
 		// remove oldTag first so that we unobserve before we re-observe
 		if (oldTag) {
 			// if there was focus, we need to figure out what element has it
-			const children = oldTag.querySelectorAll('*')
-			const parentAndChildrenNodes = [oldTag, ...children]
-			removedElementWithFocusData.index = parentAndChildrenNodes.findIndex(element => element === document.activeElement)
+			const allElements = parentAndChildrenElements(oldTag)
+			removedElementWithFocusData.index = allElements.findIndex(element => element === document.activeElement)
```
ethanjurman, line 48: Could we do a query selector here for active element, instead of checking against a large list created by a wildcard selector?

ethanjurman (approved):
Looks good, a few questions, and can you provide any examples of it failing before/ success after?

src/observe-tag.js
JRJurman: `querySelectorAll` (which we use in other places as well) will actually get all the children under an element.
image

src/observe-tag.js
```diff
@@ -56,19 +82,35 @@ module.exports = tagFunction => {
...
+					.sort(byDistanceFromIndex(removedElementWithFocusData.index))
+					.map(toNodes(allElements))
+					.find(nodeMatchesTagName)
+
+				// if the element / child exists, focus it
+				if (elementToGiveFocus !== undefined) {
+					elementToGiveFocus.focus()
+
+					// also try to set the selection, if there is a selection for this element
+					const hasSelectionStart = removedElementWithFocusData.selectionStart !== null && removedElementWithFocusData.selectionStart !== undefined
```
chtinahow, line 102: Is there a reason why you can check to see if `removedElementWithFocusData.selectionStart` is truthy? Do null and undefined come back as truthy  in this case, or is there a falsy value that you want to return as true?

src/observe-tag.js
JRJurman: The goal here is to find the index of the element, not the element itself (if that's all we needed, we could just use `document.activeElement`). I'm not sure that a more specific query selector would help... :thinking: thinking about it now, we could query selector on only things that have a positive tabIndex?

This might be just another one of those things where I want to do the performance test, and then fix as necessary.

JRJurman:
> Looks good, a few questions, and can you provide any examples of it failing before/ success after?

Summary has been updated :+1:

src/observe-tag.js
JRJurman: `selectionStart` can be `0` which is a falsy value, so yeah, we have to explicitly check if it's null or undefined.

src/observe-tag.js
JRJurman: I do agree that if there are performance issues, it's that we have to query and iterate over all the children of an element before and after. I'm... not a fan either...

chtinahow (approved):
LGTM

PR #165, opened 2021-07-18
Update and fix tests

JRJurman:
## Summary
Updated the dependencies in tram-one, and a bunch of tests started failing. It turns out the definition for `toBeVisible` now requires the element to be in the document (which in reality makes sense). Luckily, we needed to have the app in the document for the focus tests, so this was an easy fix, and we just now have the app mount in the document by default.

## Checklist
- [x] PR Summary
- [x] Tests
- [x] Version Bump


jest.config.js
```diff
@@ -1,4 +1,4 @@
 module.exports = {
 	setupFilesAfterEnv: ['@testing-library/jest-dom/extend-expect'],
-	testEnvironment: 'jest-environment-jsdom-sixteen'
+	testEnvironment: 'jsdom'
```
JRJurman, line 3: :tada:

docs/badges/umd.svg
```diff
@@ -14,7 +14,7 @@
   <g fill="#fff" text-anchor="middle" font-family="DejaVu Sans,Verdana,Geneva,sans-serif" font-size="11">
     <text x="16.5" y="15" fill="#010101" fill-opacity=".3">umd</text>
     <text x="16.5" y="14">umd</text>
-    <text x="59" y="15" fill="#010101" fill-opacity=".3">21.33 kB</text>
-    <text x="59" y="14">21.33 kB</text>
+    <text x="59" y="15" fill="#010101" fill-opacity=".3">21.41 kB</text>
+    <text x="59" y="14">21.41 kB</text>
```
JRJurman, line 18: I have no idea why this actually went up... As far as I know, the umd build shouldn't have the test logic at all :thinking:

JRJurman, line 18: this probably went up because our dependencies increased in size, or because the build tooling changed in a way where this isn't as compact. Either way, it's _fine_.

chtinahow (approved):

PR #167, opened 2021-07-19
initial work for adding performance tests

### LINKED ISSUE #39: Tram One Performance Tests
## Summary
It would be useful to have performance tests comparing Tram-One to other frameworks. It would also be good to verify that changes don't make the project any slower (without a good reason)!

Not really sure what tests we would do, this requires some research into what other frameworks are doing for performance tests.

JRJurman:
## Summary
Fixes #39 
This PR is an attempt to set up a framework for us to add tests that verify the performance of tram-one. All of the infrastructure is identical to the existing unit tests, but has uses performance timers and helper functions to get metrics of how slow tram-one was to update. It currently uses snapshots as artifacts of the timers.

## Checklist
- [x] PR Summary
- [x] Tests
- [x] Version Bump

### DIFF
--- a/performance-tests/__snapshots__/performance.test.js.snap
+++ b/performance-tests/__snapshots__/performance.test.js.snap
@@ -3,34 +3,34 @@
 exports[`Tram-One - Performance Tests should render lots of elements quickly 1`] = `
 Object {
   "0010": Object {
-    "averageTime": 1.0707592200000091,
-    "maxTime": 2.0995399999999336,
-    "medianTime": 0.9903519999998025,
+    "averageTime": 1.2213466800000423,
+    "maxTime": 2.964711999999963,
+    "medianTime": 1.0390390000002299,
   },
   "0050": Object {
-    "averageTime": 1.9795256599999538,
-    "maxTime": 3.184698999999455,
-    "medianTime": 1.8801470000003064,
+    "averageTime": 2.090679160000009,
+    "maxTime": 5.007340000000113,
+    "medianTime": 1.9029090000003634,
   },
   "0100": Object {
-    "averageTime": 3.245847500000145,
-    "maxTime": 9.136802999999418,
-    "medianTime": 2.981826000000183,
+    "averageTime": 3.4207105600000975,
+    "maxTime": 4.558337000000392,
+    "medianTime": 3.0174210000004678,
   },
   "0500": Object {
-    "averageTime": 12.792782179999922,
-    "maxTime": 17.862021000000823,
-    "medianTime": 12.381100000000515,
+    "averageTime": 13.692366439999933,
+    "maxTime": 34.392783999999665,
+    "medianTime": 13.348606999999902,
   },
   "1000": Object {
-    "averageTime": 27.069117659999684,
-    "maxTime": 36.891797000000224,
-    "medianTime": 27.212838999999803,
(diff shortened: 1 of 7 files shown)


performance-tests/performance.test.js
```diff
@@ -0,0 +1,98 @@
...
+
+	// click to trigger the render
+	userEvent.click(getByText(container, 'Render'))
+
+	// wait for the render button data to udpate
+	await waitFor(() => {
+		expect(getByText(container, 'Render')).toHaveAttribute('renders', `${renders}`)
+	})
+
+	// will be "Wait: NN.NNNNNNNN"
```
chtinahow, line 30: What does NN.NNNNNN mean? Is this a time or something else?

performance-tests/performance.test.js
```diff
@@ -0,0 +1,98 @@
...
+	it('should render lots of elements quickly', async () => {
+		// set the test to be element-rendering
+		window.history.pushState({}, '', '/element-rendering')
+
+		const { container } = startApp()
+
+		// verify that the element-rendering test is up
+		expect(container).toHaveTextContent('Element Rendering Example')
+
+		const elementCounts = ['0010', '0050', '0100', '0500', '1000', '5000']
```
chtinahow, line 62: Are these based on miliseconds?

performance-tests/test-app/element-rendering.js
```diff
@@ -0,0 +1,36 @@
...
+
+	const updateCount = event => {
+		pageStore.queue = event.target.value
+	}
+
+	const render = () => {
+		pageStore.renders++
+		pageStore.startTimer = performance.now()
+		pageStore.elements = pageStore.queue
+		pageStore.endTimer = performance.now()
```
JRJurman, line 19: these happen one at a time, which is why this whole performance test works. 

I wonder if this should be fixed though, so that all the state changes are queued and made at once, and then we rerender? I don't know if this would be technically challenging, or even possible 🤔

chtinahow (commented):
Thumb after version bump

performance-tests/performance.test.js
JRJurman: this is supposed to represent what the text we are expecting... Not sure if there's a better way to word this...  In reality it'll look something like `Wait: 12.3456789`, which is maybe just what I should write 😆

performance-tests/performance.test.js
JRJurman: these are the actual counts of elements

performance-tests/__snapshots__/performance.test.js.snap
```diff
@@ -0,0 +1,36 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`Tram-One - Performance Tests should render lots of elements quickly 1`] = `
```
JRJurman, line 3: I'm slightly worried that these numbers will change radically when running on different devices, and make it difficult to understand the numbers over time.

We could just snapshot the difference between 10 v 50, 100 v 500, 1000 v 5000, which is what we are actually expecting on in the spec, but that data would be misleading too, since a performance boost linearly would show the same ratios. We could record what the specs of the machine are, but that doesn't make clear what else is running or potentially impacting performance 🤔

PR #169, opened 2021-08-04
initial set of performance improvements

JRJurman:
## Summary

* Removes `querySelectorAll` and uses TreeWalkers instead
* names previously anonymous functions so that they appear in the performance debugger
* Fix an issue with reseting scroll position on elements that have none

Changes to improve the performance throughout the framework. See the PR comments for more details!

## Checklist
- [x] PR Summary
- [x] Tests
- [x] Version Bump

### DIFF
--- a/performance-tests/test-app/element-rendering.js
+++ b/performance-tests/test-app/element-rendering.js
@@ -1,4 +1,4 @@
-const { registerHtml, useStore, useEffect } = require('../../src/tram-one')
+const { registerHtml, useStore } = require('../../src/tram-one')
 
 const html = registerHtml()
 
--- a/performance-tests/test-app/element-rendering.js
+++ b/performance-tests/test-app/element-rendering.js
@@ -1,4 +1,4 @@
-const { registerHtml, useStore } = require('../../src/tram-one')
+const { registerHtml, useStore, useEffect } = require('../../src/tram-one')
 
 const html = registerHtml()
 
@@ -20,7 +20,8 @@ module.exports = () => {
 	}
 
 	const numberOfElements = Number.parseInt(pageStore.elements, 10)
-	const elements = [...new Array(Number.isNaN(numberOfElements) ? 0 : numberOfElements)].map(() => html`<span>-</span>`)
+	const newSpan = () => html`<span>-</span>`
+	const elements = [...new Array(Number.isNaN(numberOfElements) ? 0 : numberOfElements)].map(newSpan)
 
 	return html`
 		<section>
--- a/src/dom.js
+++ b/src/dom.js
@@ -45,10 +45,15 @@ const registerDom = (namespace, registry = {}) => {
 			}
 
 			// observe store usage and process any new effects that were called when building the component
-			const tagResult = observeTag(() => processEffects(populatedTagFunction))
+			const processEffectsAndBuildTagResult = () => processEffects(populatedTagFunction)
+			const tagResult = observeTag(processEffectsAndBuildTagResult)
 
 			// pop the branch off (since we are done rendering this component)
 			popWorkingKeyBranch(TRAM_HOOK_KEY)
+
+			// before we return the result, we need to add an attribute that will allow us to query the element
+			// this is the only visual side-effect we'll have on the DOM, all other data is on the Node
+			tagResult.setAttribute('tram', '')
 
 			return tagResult
 		}
--- a/src/dom.js
+++ b/src/dom.js
@@ -8,6 +8,7 @@ const { TRAM_HOOK_KEY } = require('./engine-names')
 const { pushWorkingKeyBranch, popWorkingKeyBranch, incrementWorkingKeyBranch, copyWorkingKey, restoreWorkingKey } = require('./working-key')
 const observeTag = require('./observe-tag')
 const processEffects = require('./process-effects')
+const { TRAM_TAG } = require('./node-names')
 
 /**
  * This function takes in a namespace and registry of custom components,
@@ -50,6 +51,9 @@ const registerDom = (namespace, registry = {}) => {
 
 			// pop the branch off (since we are done rendering this component)
 			popWorkingKeyBranch(TRAM_HOOK_KEY)
+
+			// mark this node as a tram-one component (so we can filter on it later)
+			tagResult[TRAM_TAG] = true
 
 			return tagResult
 		}
--- a/src/dom.js
+++ b/src/dom.js
@@ -51,10 +51,6 @@ const registerDom = (namespace, registry = {}) => {
 			// pop the branch off (since we are done rendering this component)
 			popWorkingKeyBranch(TRAM_HOOK_KEY)
 
-			// before we return the result, we need to add an attribute that will allow us to query the element
-			// this is the only visual side-effect we'll have on the DOM, all other data is on the Node
-			tagResult.setAttribute('tram', '')
-
 			return tagResult
 		}
 
--- a/src/mutation-observer.js
+++ b/src/mutation-observer.js
@@ -63,24 +63,36 @@ const clearNode = node => {
 	}
 }
 
+const isTramOneComponent = node => {
+	if (node.getAttribute('tram') !== null) {
+		return NodeFilter.FILTER_ACCEPT
+	}
+	return NodeFilter.FILTER_SKIP
+}
+
 // function to get the children (as a list) of the node passed in
-const childrenNodes = node => {
-	const children = node.querySelectorAll ? node.querySelectorAll('*') : []
-	return [...children]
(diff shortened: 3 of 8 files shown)


performance-tests/test-app/element-rendering.js
```diff
@@ -20,7 +20,8 @@ module.exports = () => {
 	}
 
 	const numberOfElements = Number.parseInt(pageStore.elements, 10)
-	const elements = [...new Array(Number.isNaN(numberOfElements) ? 0 : numberOfElements)].map(() => html`<span>-</span>`)
+	const newSpan = () => html`<span>-</span>`
+	const elements = [...new Array(Number.isNaN(numberOfElements) ? 0 : numberOfElements)].map(newSpan)
```
JRJurman, line 24: This is only because it makes it easier to inspect on the performance tab (otherwise it just appears as `anonymous function`

src/dom.js
```diff
@@ -45,11 +45,16 @@ const registerDom = (namespace, registry = {}) => {
 			}
 
 			// observe store usage and process any new effects that were called when building the component
-			const tagResult = observeTag(() => processEffects(populatedTagFunction))
+			const processEffectsAndBuildTagResult = () => processEffects(populatedTagFunction)
+			const tagResult = observeTag(processEffectsAndBuildTagResult)
```
JRJurman, line 49: This is only because it makes it easier to inspect on the performance tab (otherwise it just appears as `anonymous function`)

src/dom.js
```diff
@@ -45,11 +45,16 @@ const registerDom = (namespace, registry = {}) => {
...
 
 			// pop the branch off (since we are done rendering this component)
 			popWorkingKeyBranch(TRAM_HOOK_KEY)
 
+			// before we return the result, we need to add an attribute that will allow us to query the element
+			// this is the only visual side-effect we'll have on the DOM, all other data is on the Node
+			tagResult.setAttribute('tram', '')
+
```
JRJurman, line 57: This causes the components registered with Tram-One to have a `tram` attribute, see the PR description for more details.

src/mutation-observer.js
```diff
@@ -63,24 +63,38 @@ const clearNode = node => {
...
-	return [...children]
+// this only needs to query tram-one components, so we can use the attribute `tram`
+const childrenComponents = node => {
+	const componentWalker = document.createTreeWalker(node, NodeFilter.SHOW_ELEMENT, isTramOneComponent)
+	const children = []
+	while (componentWalker.nextNode()) {
+		children.push(componentWalker.currentNode)
+	}
+
+	return children
```
JRJurman, line 83: rather than using `querySelectorAll` (which forces reflow), we use a `TreeWalker`! This `TreeWalker` specifically looks for nodes that pass the `isTramOneComponent` check (which in turn, checks if the node has `TRAM_TAG_NEW_EFFECTS` - which will be some array, even empty)

src/observe-tag.js
```diff
@@ -13,10 +12,23 @@ const byDistanceFromIndex = targetIndex => (indexA, indexB) => {
...
-	const children = node.querySelectorAll('*')
-	return [node, ...children]
+const parentAndChildrenElements = (node, tagName) => {
+	const componentWalker = document.createTreeWalker(node, NodeFilter.SHOW_ELEMENT, hasMatchingTagName(tagName))
+	const parentAndChildren = [componentWalker.currentNode]
+	while (componentWalker.nextNode()) {
+		parentAndChildren.push(componentWalker.currentNode)
+	}
+
+	return parentAndChildren
```
JRJurman, line 31: Same as above, rather than using `querySelectorAll`, this iterates through all the elements using a `TreeWalker`, and saves those that have the same tagName (since those are the only ones we would reattach focus on).

src/observe-tag.js
```diff
@@ -28,37 +40,33 @@ const parentAndChildrenElements = node => {
  */
 module.exports = tagFunction => {
 	let tagResult
-	const tagReaction = observe(() => {
+	const buildAndReplaceTag = () => {
```
JRJurman, line 43: Named function so that it is easier to debug in the performance tab.

src/observe-tag.js
```diff
@@ -28,37 +40,33 @@ const parentAndChildrenElements = node => {
...
 		// remove oldTag first so that we unobserve before we re-observe
 		if (oldTag) {
-			// if there was focus, we need to figure out what element has it
-			const allElements = parentAndChildrenElements(oldTag)
-			removedElementWithFocusData.index = allElements.findIndex(element => element === document.activeElement)
+			// we need to blow away any old focus data we had
+			removedElementWithFocusData = {}
+
+			// determine if this element (or any element under it) had focus
+			const oldTagHasFocusedElement = oldTag.contains(document.activeElement)
```
JRJurman, line 54: this saves us from doing a focus check for every element!

src/observe-tag.js
```diff
@@ -78,26 +86,21 @@ module.exports = tagFunction => {
 
 		// if oldTag was defined, then we need to replace it with the new result
 		if (oldTag) {
-			oldTag.replaceWith(tagResult)
-
 			// if an element had focus, reapply it
+			let elementToGiveFocus
 			if (removedElementWithFocusData.index >= 0) {
-				const allElements = parentAndChildrenElements(tagResult)
+				const allActiveLikeElements = parentAndChildrenElements(tagResult, removedElementWithFocusData.tagName)
```
JRJurman, line 92: rather than filtering after the fact, we get only elements with the same tag name (this is why they are `active-like`

src/observe-tag.js
```diff
@@ -109,17 +112,25 @@ module.exports = tagFunction => {
 					}
 
 					// also set the scrollLeft and scrollTop (since this is reset to 0 by default)
-					elementToGiveFocus.scrollLeft = removedElementWithFocusData.scrollLeft
-					elementToGiveFocus.scrollTop = removedElementWithFocusData.scrollTop
+					if (removedElementWithFocusData.scrollLeft || removedElementWithFocusData.scrollTop) {
+						elementToGiveFocus.scrollLeft = removedElementWithFocusData.scrollLeft
+						elementToGiveFocus.scrollTop = removedElementWithFocusData.scrollTop
+					}
```
JRJurman, line 118: we verify that there is a `scrollLeft` and `scrollTop` now, previously this was causing a bug on elements that could have focus, but can't be scrolled.

src/observe-tag.js
```diff
@@ -109,17 +112,25 @@ module.exports = tagFunction => {
...
 			}
 
 			// copy the reaction and effects from the old tag to the new one
 			tagResult[TRAM_TAG_REACTION] = oldTag[TRAM_TAG_REACTION]
 			tagResult[TRAM_TAG_NEW_EFFECTS] = oldTag[TRAM_TAG_NEW_EFFECTS]
 			tagResult[TRAM_TAG_CLEANUP_EFFECTS] = oldTag[TRAM_TAG_CLEANUP_EFFECTS]
+
+			// both these actions cause forced reflow, and can be performance issues
+			oldTag.replaceWith(tagResult)
+			if (elementToGiveFocus) elementToGiveFocus.focus()
```
JRJurman, line 129: this is still causing issues, and I have no idea how to performantly fix it. We need to replace before we focus, because you can't focus an element that isn't in the document. If there is a way we can batch it, that would be magic, but I haven't found anything that offers that (yet).
image

src/observe-tag.js
chtinahow: Weird wording question. scrollLeft makes sense, but scrollTop seems a bit weird. scrollToTop sounds better?

src/observe-tag.js
```diff
@@ -13,10 +12,23 @@ const byDistanceFromIndex = targetIndex => (indexA, indexB) => {
 	return diffFromTargetA - diffFromTargetB
 }
 
+const hasMatchingTagName = tagName => node => {
+	if (node.tagName === tagName) {
+		return NodeFilter.FILTER_ACCEPT
+	}
```
chtinahow, line 18: You have a one liner here, but you wrap it in curly braces, but lower in the code, you do the one line return sans curly. I would keep it consistent

ethanjurman (approved):
Max time in 1000+ elements 😭
Everything else is a huge boost so yay

src/observe-tag.js
JRJurman: Those are built into the DOM, so we can't really pick and choose those :man_shrugging:

src/observe-tag.js
JRJurman: I've asked this question on stack-overflow, so we'll see if we get a response - https://stackoverflow.com/questions/68675817/how-can-i-group-javascript-actions-that-force-reflow

PR #170, opened 2021-08-06
Test Performance on CI

JRJurman:
## Summary

This PR adds the configuration to run the performance tests on CircleCI, makes them a little more consistent, and pushes the performance tests (and build badges!) back to the repo.

* When working on a non-master branch, this creates a new branch with the `-artifacts` suffix.
  * These branches never have to be merged (we will have to clean them up later, but that's fine for now)
* When working on master (after a PR has been merged), this creates a new commit straight on master
  * This will include the final performance result and badge result, regardless of whatever was committed in the PR

## Performance / Size
Performance Results - https://github.com/Tram-One/tram-one/compare/performance-test-on-ci...performance-test-on-ci-artifacts

## Resources
* This followed the following guide to setup the pushing from CircleCI - https://blog.jdblischak.com/posts/circleci-ssh/

## Checklist
- [x] PR Summary
- [x] Tests
- [x] Version Bump

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -26,6 +26,11 @@ jobs:
           # fallback to using the latest cache if no exact match is found
           # - v1-dependencies-
 
+      # Add ssh key so we can publish build assets to the repo
+      - add_ssh_keys:
+          fingerprints:
+            - "64:3e:d4:d8:4b:95:68:79:d3:3b:ab:b1:5c:fa:2d:3d"
+
       - run: npm install
 
       - save_cache:
@@ -40,7 +45,27 @@ jobs:
       - store_artifacts:
           path: coverage/lcov-report/
 
+      - run: npm run test:performance:ci
+
+      - store_artifacts:
+          path: performance-tests/__snapshots__
+
       - run: npm run build
 
       - store_artifacts:
           path: dist
+
+      - run:
+          name: Commit Assets
+          command: |
+            if [[ -z "${CIRCLE_PULL_REQUEST}" ]]
+            then
+              echo "Master Build, not committing to GitHub"
+            else
+              echo "PR Build, Committing to GitHub"
+              git config user.email "j.r.jurman@gmail.com"
+              git config user.name "CircleCI Job"
+              git checkout
+              git commit --allow-empty -am "Update Snapshots & Badges [skip ci]"
+              git push -u origin HEAD
+            fi
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -58,14 +58,17 @@ jobs:
       - run:
           name: Commit Assets
           command: |
+            # configuring user
+            git config user.email "j.r.jurman@gmail.com"
+            git config user.name "CircleCI Job"
+
             if [[ -z "${CIRCLE_PULL_REQUEST}" ]]
             then
-              echo "Master Build, not committing to GitHub"
-            else
-              echo "PR Build, Committing to GitHub"
-              git config user.email "j.r.jurman@gmail.com"
-              git config user.name "CircleCI Job"
+              echo "Master Build, checkout to master"
               git checkout
-              git commit --allow-empty -am "Update Snapshots & Badges [skip ci]"
-              git push -u origin HEAD
+            else
+              echo "PR Build, checkout to new branch"
+              git checkout -b ${CIRCLE_BRANCH}-artifacts
             fi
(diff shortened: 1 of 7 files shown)


src/tram-one.js
```diff
@@ -10,7 +10,7 @@ const start = require('./start')
  * @package
  * @public
  * @description
- * Tram-One is a light View Framework that comes with all the dependencies you need to start developing on the web.
+ * Tram-One is a Modern View Framework that has advance features like hooks, observables, and JSX-like template components, all in plain vanilla javascript.
```
JRJurman, line 13: needed something to test that the badge sizes would be updated, this worked, and this needed to be updated anyways.

performance-tests/performance.test.js
```diff
@@ -93,8 +106,8 @@ describe('Tram-One - Performance Tests', () => {
 		// we'll use the median, since that is the least prone to error from outliers
 		// we'll also add a buffer to the right, to account for fragility
 		const buffer = 1 // 1 seconds, which will map to 5 seconds on the left
-		expect(stats['0050'].medianTime / 5).not.toBeGreaterThan(stats['0010'].medianTime + buffer)
-		expect(stats['0500'].medianTime / 5).not.toBeGreaterThan(stats['0100'].medianTime + buffer)
-		expect(stats['5000'].medianTime / 5).not.toBeGreaterThan(stats['1000'].medianTime + buffer)
+		expect(stats['0050'].averageTime / 5).not.toBeGreaterThan(stats['0010'].averageTime + buffer)
+		expect(stats['0500'].averageTime / 5).not.toBeGreaterThan(stats['0100'].averageTime + buffer)
+		expect(stats['5000'].averageTime / 5).not.toBeGreaterThan(stats['1000'].averageTime + buffer)
```
JRJurman, line 111: using `averageTime` instead of `medianTime`, since that should be the least error prone (with the outliers removed)

performance-tests/performance.test.js
```diff
@@ -5,6 +5,11 @@ const { getByText, getByLabelText, waitFor } = require('@testing-library/dom')
 const { default: userEvent } = require('@testing-library/user-event')
 const { startApp } = require('./test-app')
 
+// number of times to run the test
+const NUMBER_OF_RUNS = 50
+// the number of tests to toss (slowest and fastest)
+const BUFFER = 10
```
JRJurman, line 11: including a buffer now, which is the number of slowest (and fastest) runs that we'll toss when doing calculations. In reality, this means that we are only getting stats for 30 runs (50 runs, - 10 slowest, - 10 fastest), but these should be more consistent overall.

.circleci/config.yml
```diff
@@ -40,7 +45,30 @@ jobs:
       - store_artifacts:
           path: coverage/lcov-report/
 
+      - run: npm run test:performance:ci
+
+      - store_artifacts:
+          path: performance-tests/__snapshots__
```
JRJurman, line 51: while not totally required, this does give us a historical view within a branch (since when we make our `-artifacts` branch, we always just force-push over the commit)

.circleci/config.yml
```diff
@@ -40,7 +45,30 @@ jobs:
...
+            git config user.name "CircleCI Job"
+
+            if [[ -z "${CIRCLE_PULL_REQUEST}" ]]
+            then
+              echo "Master Build, checkout to master"
+              git checkout
+            else
+              echo "PR Build, checkout to new branch"
+              git checkout -b ${CIRCLE_BRANCH}-artifacts
+            fi
```
JRJurman, line 72: Some configurations only commit when on the master branch, but since we want information on PRs as well, this is allows us to account for both :+1:

.circleci/config.yml
```diff
@@ -40,7 +45,30 @@ jobs:
...
+            git commit --allow-empty -am "Update Snapshots & Badges [skip ci]"
```
JRJurman, line 73: `[skip ci]` is important here, this actually prevents CircleCI from infinitely triggering builds.

.circleci/config.yml
```diff
@@ -40,7 +45,30 @@ jobs:
...
+            git push -fu origin HEAD
```
JRJurman, line 74: we force push the artifacts onto the branch, this is required for the Pull Request use case, less so for master, but I don't foresee that being an actual issue.

chtinahow (approved):
👍

PR #171, opened 2021-08-07
Tag and Hook Warnings

### LINKED ISSUE #158: Warn when a component doesn't return DOM (or doesn't return anything)
## Summary
Currently, when a component doesn't return anything, Tram-One fails trying to process effects.
We should have logging if there is no tagResult here (or anywhere else where we try to make a tag, this might be the only place though)

https://github.com/Tram-One/tram-one/blob/15536a5801ad32b2a03cc570268c1047f4dab4c3/src/process-effects.js#L19

JRJurman:
## Summary

This PR adds warnings when the tram-one API is used incorrectly.
* if you forget to return anything (or an error would cause a component to not return anything)
  * fixes #158
* if you call a hook outside of a component (before the app has a chance to start) 

## Artifacts
https://github.com/Tram-One/tram-one/compare/tag-warning...tag-warning-artifacts

## Checklist
- [x] PR Summary
- [x] Tests
- [x] Version Bump


JRJurman:
(will do a version bump after other PR is merged in... in reality, that should also be controlled by the CI)

configs/rollup.config.umd.js
```diff
@@ -23,7 +23,7 @@ const plugins = [
 ]
 
 // domino is a package used by belit to support server side rendering,
-// it does not need to be included in browser buidls, which will have document
+// it does not need to be included in browser builds, which will have document
```
JRJurman, line 26: saw this typo, wanted to fix it.

integration-tests/broken-app/index.js
```diff
@@ -0,0 +1,37 @@
...
+	}
+}
+
+if (document.querySelector('#parcel-page')) {
+	startApp('#parcel-page')
+}
+
+module.exports = {
+	app, startApp
+}
```
JRJurman, line 37: This is copied-pasted a lot, I should make this a helper function or something (if exports can be put in helper functions? maybe just the `startApp` :thinking: )

integration-tests/broken-hook/title.js
```diff
@@ -0,0 +1,18 @@
+const { registerHtml, useEffect } = require('../../src/tram-one')
+
+const html = registerHtml()
+
+// other hooks also generate the same error
+// so we don't need to check those, just useEffect is fine
+useEffect(() => {})
```
JRJurman, line 7: validated manually that other hooks also generate the same error.

integration-tests/integration.test.js
```diff
@@ -1,7 +1,13 @@
 const { getByText, fireEvent, waitFor } = require('@testing-library/dom')
 const { startApp } = require('./test-app')
+const { startApp: startBrokenApp } = require('./broken-app')
 
 describe('Tram-One', () => {
+	beforeEach(() => {
+		// clean up any tram-one properties between tests
+		window['tram-space'] = undefined
+	})
```
JRJurman, line 9: I need to do this for the hook test (because that's how we know that a hook was called in the wrong context.

performance-tests/performance.test.js
```diff
@@ -105,7 +105,7 @@ describe('Tram-One - Performance Tests', () => {
 		// what does quickly mean? 5x elements should not be more than 5x slower
 		// we'll use the average, since that should be the least error prone (with the outliers removed)
 		// we'll also add a buffer to the right, to account for fragility
-		const buffer = 1 // 1 seconds, which will map to 5 seconds on the left
+		const buffer = 2 // 2 seconds, which will map to 10 seconds on the left
```
JRJurman, line 108: this test was failing intermittently on CircleCI, so this should fix it (in reality, a difference of 10 seconds is more than fine.

JRJurman: I guess, it does matter for the small cases though, maybe it should be a percentage of the average time :thinking:

src/mutation-observer.js
```diff
@@ -107,4 +107,4 @@ const startWatcher = (observerName, node) => {
 	observerStore.observe(node, { childList: true, subtree: true })
 }
 
-module.exports = { setupMutationObserver, getMutationObserver, startWatcher }
+module.exports = { setupMutationObserver, startWatcher }
```
JRJurman, line 110: removed unused export (this was determined manually)

src/working-key.js
```diff
@@ -93,8 +93,6 @@ module.exports = {
 	pushWorkingKeyBranch,
 	popWorkingKeyBranch,
 	incrementWorkingKeyBranch,
-	getWorkingKey,
-	getWorkingBranch,
```
JRJurman, line 97: removed unused export (this was determined manually)

src/namespace.js
```diff
@@ -15,6 +15,14 @@ const setup = constructor => {
...
 
 const get = namespace => {
+	// if tram-one is setup, this will be defined
+	const tramOneIsSetup = window['tram-space']
+
+	// otherwise, we should warn
+	if (!tramOneIsSetup) {
+		throw new Error('Tram-One: app has not started yet, but hook was called. Is it being invoked outside a component function?')
+	}
+
```
JRJurman, line 25: Error to throw when `window['tram-space']` is not setup. 
This happens most often when a hook is called outside of a component function (it's called immediately when a component is imported, and thus, before start has been called, and before the `window['tram-space']` has been defined.). It previously returned the following error:
```
"Cannot read property 'tram-effect-queue' of undefined"
```
But this should be more clear to developers, and not require them to understand the inner-workers (or know how to parse this somewhat cryptic error).

src/process-effects.js
```diff
@@ -17,6 +17,9 @@ module.exports = tagFunction => {
 
 	// create the component, which will save new effects to the effect queue
 	const tagResult = tagFunction()
+	if (tagResult === undefined) {
+		throw new Error('Tram-One: no element returned from tag, is anything being returned?')
+	}
```
JRJurman, line 22: Error to throw when a component does not return an element.
Previously, this would throw the following error, because no DOM object was being returned.
```
"Cannot set property 'tram-tag-new-effects' of undefined"
```

src/process-effects.js
```diff
@@ -18,6 +18,19 @@ module.exports = tagFunction => {
 	// create the component, which will save new effects to the effect queue
 	const tagResult = tagFunction()
 
+	if (tagResult instanceof Array) {
+		throw new Error('Tram-One: Sorry, Tram-One does not currently support array returns. Wrap components in an element before returning.')
+	}
```
JRJurman, line 23: This could be eventually resolved by #173 , for now, an error should notify people that support doesn't exist (yet).

src/process-effects.js
```diff
@@ -18,6 +18,15 @@ module.exports = tagFunction => {
...
+
+	// verify that the tagResult is an element (if it's not, we won't be able to run effects or do anything useful)
+	if (!(tagResult instanceof Element)) {
+		throw new Error(`Tram-One: expected component to return an Element, instead got ${typeof tagResult}. Verify the component is a function that returns DOM.`)
+	}
```
JRJurman, line 28: Error to throw when a component does not return an element.
Previously, this would throw the following error, because no DOM object was being returned.
```
"Cannot set property 'tram-tag-new-effects' of undefined"
```
If the thing being returned was not an element, other issues would also occur (most notably effects would not process).

Now we verify that something is returned, and that it's an Element.

src/process-effects.js
```diff
@@ -18,6 +18,15 @@ module.exports = tagFunction => {
 	// create the component, which will save new effects to the effect queue
 	const tagResult = tagFunction()
 
+	if (Array.isArray(tagResult)) {
+		throw new TypeError('Tram-One: Sorry, Tram-One does not currently support array returns. Wrap components in an element before returning.')
+	}
```
JRJurman, line 23: This could be eventually resolved by #173 , for now, an error should notify people that support doesn't exist (yet).

src/process-effects.js
```diff
@@ -18,6 +18,15 @@ module.exports = tagFunction => {
...
+
+	// verify that the tagResult is an element (if it's not, we won't be able to run effects or do anything useful)
+	if (!(tagResult instanceof Element)) {
+		throw new TypeError(`Tram-One: expected component to return an Element, instead got ${typeof tagResult}. Verify the component is a function that returns DOM.`)
+	}
```
JRJurman, line 28: Error to throw when a component does not return an element.
Previously, this would throw the following error, because no DOM object was being returned.
```
"Cannot set property 'tram-tag-new-effects' of undefined"
```
If the thing being returned was not an element, other issues would also occur (most notably effects would not process).

Now we verify that something is returned, and that it's an Element.

chtinahow (approved):

PR #175, opened 2021-10-21
Tram-One, now with Types! Typescript! v11.0.0

JRJurman:
## Summary
In an effort to get the advantages of type safety, this PR migrates the source code (and most of the tests) to Typescript. 

This gives the project more type safety, and allows our consumers to build projects with type safety (if they are using Typescript). If they aren't using Typescript they still will get proper type annotations when working with compatible editors (such as Visual Studio Code).

## Changes
All the changes listed below are detailed in the comments, treat these as a high level breakdown.

### Bump Major - v11.0.0
In reality, this might not be breaking for JS projects, but there are a significant number of changes (including the removal of some safety checks for JS and addition of checks for TS). Anyone using Typescript would start getting compile time errors / warnings, so for that reason alone really it should be considered a breaking change.

### Main Changes
* Remove manual type checking in the project
* Remove manual JS Doc annotations
* `types.ts` file for typescript type definitions

### Auxiliary Changes
* Using Prettier instead XO
* Expand `...args` for more clear parameter definitions
* Fix file JSDoc blocks to not be associated with a variable
* Fixed Document TreeWalker to properly filter elements
* Removed server-side rendering support

### Testing Changes
* Added new test for verifying root-defined effects
* Created #176 (we lost code coverage for elements that have scroll position, in reality, we probably never actually had coverage for these, and it was being ignored)

Artifacts: https://github.com/Tram-One/tram-one/compare/typescript...typescript-artifacts
<details>
<summary>Note about the artifacts</summary>
For some reason, CircleCI slowed down between the last master run and this changeset. This was confirmed with a dummy PR which had similar slowdowns (with no changes).

</details>

## Checklist
- [x] PR Summary
- [x] Update tsconfig to be strict
- [x] Update DocStrings
- [x] Verify DocStrings
- [x] Update Build tooling for Modules
- [x] Manual Verification in a JS Project
- [x] Manual Verification in a TS Project
- [x] Investigate slower speeds (run dummy PR against master)
- [x] Update top-level README
- [x] PR Comments (Initial Pass)
- [x] PR Comments (Final Pass)
- [x] Update Linter (now including prettier)
- [x] Tests
- [x] Version Bump
- [x] Code Coverage Review

## Post Merge Checklist
- [ ] Update Website


.gitignore
```diff
@@ -4,20 +4,8 @@ dist
...
-logs
-*.log
-*.log.*
-
 # Dependency directory
 node_modules
 
-# SSL certificates
-.ssl
-
```
JRJurman, line 20: This was most certainly an artifact of when we had the website hosted from this repo, since that is no longer a thing, we can remove this

.prettierrc.js
```diff
@@ -0,0 +1,5 @@
+module.exports = {
+	useTabs: true,
+	singleQuote: true,
+	printWidth: 120,
+};
```
JRJurman, line 5: Using prettier instead of xo. 

The prettier config is more known to work with typescript, and xo was breaking too much.

Eventually I'd like to rip out the eslint config for tram-one projects as it is today. It really should be something that validates tram-one things (e.g., I would love to lint template spacing, and hook placement). Right now, it's just silly opinions, and that doesn't really help anyone.

JRJurman: This includes using semicolons (which sadly blows the number of lines changed for the project super high). ASI has burned me enough times (read, more than 0 times), and I think with the new Typescript syntax, it's easier if we are on the same standard as everyone else.

integration-tests/test-app/sub-mirror-input.ts
```diff
@@ -0,0 +1,34 @@
...
+	'sub-mirrorable-input': require('./sub-mirror-input'),
+});
+
+type InputObject = { value: string };
+
+/**
+ * component to test url parameters
+ */
+export default () => {
+	const mirrorable = useGlobalStore('mirrorable-input') as InputObject;
```
JRJurman, line 13: The type isn't actually required if you pass in a default value. The fact that this option exists is cool, and possible because of function overloading.

integration-tests/warnings.test.js
```diff
@@ -1,29 +1,25 @@
...
-		expect(() => startBrokenApp('empty')()).toThrowError('Tram-One: expected component to return an Element, instead got undefined. Verify the component is a function that returns DOM.')
-	})
-
-	it('should warn if a component does not return an element', () => {
-		expect(() => startBrokenApp('non-dom')()).toThrowError('Tram-One: expected component to return an Element, instead got string. Verify the component is a function that returns DOM.')
-	})
-
-	it('should warn if a component returns an array', () => {
-		expect(() => startBrokenApp('array')()).toThrowError('Tram-One: Sorry, Tram-One does not currently support array returns. Wrap components in an element before returning.')
-	})
```
JRJurman, line 24: aforementioned removed tests

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "tram-one",
-  "version": "10.1.11",
+  "version": "11.0.0",
```
JRJurman, line 3: Major bump. In reality, this might not be breaking for JS projects, but anyone using Typescript would start getting compile time errors / warnings, so for that reason, it should be breaking.

src/build-mounting-container.ts
```diff
@@ -0,0 +1,40 @@
+import { ElementOrSelector } from './types';
+
+/**
+ * Helper function for getting an element when given a string or element
+ * @param target a selector or the container
+ * @returns the container that we can mount on
+ */
+const getContainer = (target: ElementOrSelector): HTMLElement => {
```
JRJurman, line 8: DocStrings no longer have types in them. These _should_ be inherited from the Types in the method definition / implementation.

src/build-mounting-container.ts
```diff
@@ -0,0 +1,40 @@
...
+};
+
+/**
+ * Function to determine (or create) the element that we will mount our tram-one app onto
+ * @param target either a CSS selector, or HTMLElement to attach the component to.
+ * This elememnt should be initially empty.
+ *
+ * @returns the container, now with a div that tram-one can manage
+ */
+export default (target: ElementOrSelector): HTMLElement => {
```
JRJurman, line 29: DocStrings no longer have types in them. These _should_ be inherited from the Types in the method definition / implementation.

src/dom.ts
```diff
@@ -0,0 +1,68 @@
...
+			incrementWorkingKeyBranch(TRAM_HOOK_KEY);
+			const uniqueBranch = copyWorkingKey(TRAM_HOOK_KEY);
+
+			// create a tag function that has the args passed in
+			const populatedTagFunction = () => {
+				// reset working key so we have the correct place when starting a new component
+				restoreWorkingKey(TRAM_HOOK_KEY, uniqueBranch);
+
+				return tagFunction(props, children);
+			};
```
JRJurman, line 49: Previously this function used `(...args)` instead of `(props, children)` in method definitions. Potentially at the time I didn't know what this method definition would be, or wanted it to be more flexible, or maybe it was really _really_ early on when component methods had a ton of things in them. Either way, it looks like this now, and should be easier to understand and more explicit.

src/effect-store.ts
```diff
@@ -1,33 +1,27 @@
-/**
+/*
```
JRJurman, line 1: In some places I had inadvertently (or lazily) use `/**` which makes blocks more readily, but gets consumed as docstrings for the next variable, which was not the intention.

src/namespace.ts
```diff
@@ -0,0 +1,37 @@
+/*
+ * namespace is a generic interface for global tram-one state that needs
+ * to be persisted in the app container. It exposes a setup and get function.
+ */
+
+export const setupTramOneSpace = () => {
+	window['tram-space'] = {};
+};
+
+export const buildNamespace = <NamespaceStore>(constructor: () => NamespaceStore) => {
```
JRJurman, line 10: In order to have the `get` `set` and `setup` functions all produce the same Type, we now have one generic function whose type is inferred by the constructor that is used.

src/observable-hook.ts
```diff
@@ -0,0 +1,37 @@
...
+import { StoreObject } from './types';
+
+/**
+ * Shared source code for both observable hooks, useStore, and useGlobalStore.
+ * This hook exposes a globally stored value (in either case), that can cause the component
+ * to update when a subfield of that value is updated.
+ *
+ * It has a similar interface to React's useState
+ */
+export default <Store extends StoreObject>(key?: string, value?: Store): Store => {
```
JRJurman, line 14: Previously this function verified the type of the string and value, now that can be verified with Typescript! JS users will still get annotations in Typescript capable editors.

src/observe-tag.ts
```diff
@@ -35,102 +51,108 @@ const parentAndChildrenElements = (node, tagName) => {
...
 
 			// determine if this element (or any element under it) had focus
-			const oldTagHasFocusedElement = oldTag.contains(document.activeElement)
+			const oldTagHasFocusedElement = oldTag.contains(document.activeElement);
 
 			// if an element had focus, copy over all the selection data (so we can copy it back later)
 			if (oldTagHasFocusedElement) {
+				// we'll assume that the element is an HTMLInputElement, in reality other kinds of elements will be caught here,
+				// but that's fine, since they have null as selection attributes, and setting them to null is fine
+				const activeElement = document.activeElement as HTMLInputElement;
```
JRJurman, line 73: Pretty much what it says in the comment. This is because Typescript (and somewhat inaccurately so) complains that "Element" doesn't have the selection attributes (e.g. `selectionStart`), in reality, these will be null, which I guess for Typescript is just as bad, but for us, that is fine.

src/process-effects.ts
```diff
@@ -0,0 +1,34 @@
...
+import { getEffectStore, clearEffectStore, restoreEffectStore } from './effect-store';
+import { TramOneComponent } from './types';
+
+/**
+ * This is a helper function for the dom creation.
+ * This function stores any effects generated when building a tag in resulting node that is generated.
+ *
+ * These are later processed by the mutation-observer, and cleaned up when the node is removed by the mutation-observer.
+ */
+export default (tagFunction: TramOneComponent) => {
```
JRJurman, line 12: Previously this function verified that the function returned an Element, but now we get that with Typescript!

src/types.ts
```diff
@@ -0,0 +1,68 @@
+/**
+ * Type for when we can take a CSS Selector, or an HTML Element (mostly mounting).
+ */
+export type ElementOrSelector = keyof HTMLElementTagNameMap | HTMLElement;
```
JRJurman, line 4: This file contains all the types used in the project. There are places where type definitions will be inferred, or would not be trivial to add here. Where possible I've included them here, and will probably add more as they become necessary.

JRJurman:
When playing around in a typescript project, people can use the following syntax to verify the app works as expected:
```typescript
import { registerHtml, TramOneComponent } from 'tram-one'

const html = registerHtml()

const AppHeader: TramOneComponent  = (props, children) => {
	return html`
		<h1 class="app-header">
			${children}
		</h1>
	`
}

export default AppHeader
```

docs/runkit_example.js
```diff
@@ -1,13 +1,17 @@
-const { registerHtml } = require('tram-one')
+// NOTE: This won't work since Tram-One does not (right now) support server-side rendering
+// This is on our backlog though, and if it is something that you would like, please vote / comment here:
+// https://github.com/Tram-One/tram-one/issues/181
```
JRJurman, line 3: Domino was removed, so this runkit example won't work, but I'm keeping it here because one day it hopefully will :smile:

JRJurman, line 3: Domino was removed, so this runkit example won't work, but I'm keeping it here because one day it hopefully will :smile:

integration-tests/test-app/account.ts
```diff
@@ -0,0 +1,18 @@
+import { registerHtml, useUrlParams, TramOneComponent } from '../../src/tram-one';
+
+const html = registerHtml();
+
+/**
+ * component to test url parameters
+ */
+const account: TramOneComponent = () => {
```
JRJurman, line 8: This is a pattern components can use, to verify that they follow the interface for a "TramOneComponent"

JRJurman, line 8: This is a pattern components can use, to verify that they follow the interface for a "TramOneComponent"

load-package.cjs
```diff
@@ -0,0 +1,3 @@
+// file needed for importing package-json
+// see https://rollupjs.org/guide/en/#using-untranspiled-config-files
+module.exports = require('./package.json');
```
JRJurman, line 3: basically what the comments say, this is required for loading the package.json in es modules.

package.json
```diff
@@ -1,99 +1,100 @@
...
+	"umd": "dist/tram-one.js",
+	"module": "dist/tram-one.mjs",
+	"types": "dist/tram-one.d.ts",
+	"type": "module",
+	"exports": {
+		".": {
+			"import": "./dist/tram-one.mjs",
+			"require": "./dist/tram-one.cjs"
+		}
+	},
```
JRJurman, line 16: using types and export, this may not be totally consumed correctly until the next version of typescript, but we can future proof it :+1:

src/dom.ts
```diff
@@ -0,0 +1,71 @@
...
+			const tagResult = observeTag(processEffectsAndBuildTagResult);
+
+			// pop the branch off (since we are done rendering this component)
+			popWorkingKeyBranch(TRAM_HOOK_KEY);
+
+			// decorate the properties expected on TramOneElements (see node-names.ts)
+			tagResult[TRAM_TAG] = true;
+			// we won't decorate TRAM_TAG_REACTION, that needs to be done later when we observe the tag
+			tagResult[TRAM_TAG_NEW_EFFECTS] = tagResult[TRAM_TAG_NEW_EFFECTS] || [];
+			tagResult[TRAM_TAG_CLEANUP_EFFECTS] = tagResult[TRAM_TAG_NEW_EFFECTS] || [];
```
JRJurman, line 62: decorating all the properties of a tram one tag, so that they are available for sure

src/tram-one.ts
```diff
@@ -0,0 +1,17 @@
...
+	ElementOrSelector,
+	DOMTaggedTemplateFunction,
+	TramOneComponent,
+	StoreObject,
+	CleanupEffect,
+	Effect,
+	Props,
+	Registry,
+	UrlMatchResults,
+} from './types';
```
JRJurman, line 17: Exposing these types publicly, so that other Typescript projects can use them directly.

src/types.ts
```diff
@@ -0,0 +1,153 @@
...
+/* ============= PUBLIC TYPES ========================================
+ * A lot of the types here are wrapped using an array / index of 0.
+ * This allows us to expose the types as readable on the consumption side.
+ * =====================================================================
+ */
+
+/**
+ * Type for when we can take a CSS Selector, or an HTML Element (mostly mounting).
+ */
+export type ElementOrSelector = [string | HTMLElement][0];
```
JRJurman, line 12: I'm not 100% sure about this pattern, but it allows consumers to understand what the library is looking for (rather than just saying "ElementOrSelector")

ethanjurman (approved):
Awesome stuff! Cannot wait for typescript magic!!

chtinahow (approved):
👍

PR #183, opened 2022-03-19
Fix bug where first resolved element effects are not processed

JRJurman:
## Summary
While trying to verify other performance issues, this bug was found where if the first element resolved had a useEffect, then it would not be processed because the `EFFECT_STORE` and `EFFECT_QUEUE` were the same object (and so trying to determine what new effects existed was thrown off).

## Checklist
- [x] PR Summary
- [x] Tests
- [x] Version Bump


src/effect-store.ts
```diff
@@ -7,16 +7,17 @@
...
 import { buildNamespace } from './namespace';
-import { Effect } from './types';
+import { EffectStore } from './types';
 
-const defaultEffectStore = {} as {
-	[callLikeKey: string]: Effect;
+const newDefaultEffectStore = () => {
+	return {} as EffectStore;
 };
+
```
JRJurman, line 15: this is the fix - by making this a function that generates an object, rather than defaulted as an object, we ensure that new objects are created when we create the effect queue and store.

integration-tests/test-app/index.ts
```diff
@@ -32,6 +34,9 @@ export const app = () => {
 
 	return html`
 		<main>
+			<!-- for regression testing, it is important that document-title-setter is the first element -->
+			<document-title-setter />
```
chtinahow, line 38: Why is it important that the document-title-setter is the first element? What would happen if you needed some sort of error boundary before this? Or is it static?

JRJurman: as long as it is the first resolved element (specifically the first element to have its effects resolved, which depends on it be the first leaf-node), then the tests will work as expected - the issue is if this was ever moved, the tests would erroneously pass.

chtinahow (approved):
👍  LGTM

PR #187, opened 2022-05-09
Clean up Unused Observable Stores (v12.0.0)

### LINKED ISSUE #184: Range input causes error trying to re-attach focus
## Summary
When using an input with type range, trying to re-attach focus (and specifically set selectionStart / selectionEnd / selectionDirection) causes an error.

## Potential Solution
We should wrap this logic in a try-catch. Usually I would suggest to do a more complex check here, but this happens when the element already has a setSelectionRange, so there isn't much else to check as far as I can tell.

### LINKED ISSUE #185: Clean up stores from elements that have been removed
## Summary
After building https://redditcomber.com/, a memory leak was discovered to slow down the app over successive interactions. 

After lots of investigation, it appears that one source of the memory leak could be that stores are not cleaned up if an element is removed.

We should update the logic to clean up stores for elements that have been removed.

JRJurman:
## Summary
As a side-effect of working on [RedditComber](https://redditcomber.com/) a considerable performance slow down was found when interacting with the app for a period of time. This was determined to be a memory leak, caused by local stores remaining in the app even when those elements were removed from the page.

This PR fixes #185 (and additionally fixes #184)

**For anyone that wants a demonstration / detailed walk-through of this PR, feel free to reach out to me on the Tram-One Discord**

------------

## Changes
All the changes listed below are detailed in the comments, treat these as a high level breakdown.

### Bump Major - v12.0.0
Because we now blow away state when an element is removed, this could be considered as a breaking change. While it was never intended that this state would persist in this way, it could certainly be seen as a feature. In reality, it makes more sense for people to use `useGlobalStore` if they want this persistent state.

If it turns out that this feature was super useful, we could consider making a new state hook, that was local to the component, but never blown away.

### Main Changes

* Introduced a new `KeyStore` and `KeyQueue`, which keeps track of store keys associated with elements
* Blow away local generated by `useStore` when an element is removed

### Auxiliary Changes

* Fix a bug where non-standard `input` types (e.g. range) would break when trying to reattach focus. 

### Testing Changes

* introduce an input control (which uncovered the range-input issue), that starts off like this:
image
and then can be changed to create or remove stores (each button is an element with a local count):
image

### Performance / Size Changes
See https://github.com/Tram-One/tram-one/compare/range-and-leaks...range-and-leaks-artifacts
The TLDR is that the size increases by a small amount, and the performance test is just ever so slightly slower, but not significantly.

This may seem counter-intuitive given that this PR was intended to remove performance issues, but the nature of the existing performance tests don't stress this kind of behavior.

## Todo
- [x] Verify with redditcomber.com
- [x] Annotate PR

## Checklist
- [x] PR Summary
- [x] Tests
- [x] Version Bump


integration-tests/integration.test.js
```diff
@@ -1,12 +1,11 @@
 const { getByText, fireEvent, waitFor } = require('@testing-library/dom');
 const { startApp } = require('./test-app');
 
-describe('Tram-One', () => {
-	beforeEach(() => {
-		// clean up any tram-one properties between tests
-		window['tram-space'] = undefined;
-	});
```
JRJurman, line 8: TLDR, I didn't notice this existed, ended up wasting a lot of time trying to reason around the global 'tram-space' in the other test files, and ended up making this clean up part of the `startApp` method.

integration-tests/internals.test.js
```diff
@@ -0,0 +1,70 @@
...
+			// check that the lists are the same (they may have shuffled, so sort them)
+			expect(postChangeStoresTwo.sort()).toEqual(initialStores.sort());
+		});
+	});
+
+	it('should not have recursive working-key branches', async () => {
+		// start the app
+		await startAppAndWait();
+
+		// previously the working branch indices would have long recursive chains of branches
```
JRJurman, line 56: in reality, this was a side-effect of not cleaning up the `tram-space` in the unit tests, but I was concerned so I ended up writing this test.

I believe this is also happening when hot-reloading occurs - I haven't noticed any significant side-effects from that, but we should look into it...

JRJurman: Issue Created: #188

integration-tests/regression.test.js
```diff
@@ -1,11 +1,16 @@
-const { getByText, queryAllByText, fireEvent, waitFor, getByLabelText } = require('@testing-library/dom');
+const { getByText, queryAllByText, fireEvent, waitFor, getByLabelText, queryByText } = require('@testing-library/dom');
 const { default: userEvent } = require('@testing-library/user-event');
-const { startApp } = require('./test-app');
+const { startAppAndWait } = require('./test-helpers');
```
JRJurman, line 3: Using `startAppAndWait` here instead of `startApp` - while not all tests need this, I'd like to slowly start making this the standard. See `test-helpers.ts` to understand what this does.

integration-tests/regression.test.js
```diff
@@ -89,7 +94,7 @@ describe('Tram-One', () => {
 		});
 
 		// clear the input
-		userEvent.type(getByLabelText(container, 'New Task Label'), '{selectall}{backspace}');
+		userEvent.clear(getByLabelText(container, 'New Task Label'));
```
JRJurman, line 97: didn't realize that this `clear` method existed, but yeah, it very specifically says it [selects the text and then deletes it](https://testing-library.com/docs/ecosystem-user-event/#clearelement)

integration-tests/internals.test.js
```diff
@@ -0,0 +1,70 @@
...
+const { default: userEvent } = require('@testing-library/user-event');
+const { startAppAndWait } = require('./test-helpers');
+
+/**
+ * The following suite of tests verify the behavior of the internals of Tram-One, more so than other tests might.
+ * They are often inpercievable to end-users, and verify the expected behavior of the behind-the-scenes design.
+ */
+
+describe('Tram-One', () => {
+	it('should clean up stores for elements that are no longer rendered', async () => {
```
JRJurman, line 11: this is the more nitty-gritty, internals-aware version a test that we have in the regressions suite. This was more of a stepping stone for those other tests, and while slightly redundant, it acts as a good sanity check if those other tests end up failing. 

In reality, if the internals of the app change, this could end up needing to be scrapped, but we'll cross that bridge when we get there.

integration-tests/regression.test.js
```diff
@@ -232,11 +237,140 @@ describe('Tram-One', () => {
...
+		// focus on the input (the range input defaults to 0)
+		userEvent.click(getByLabelText(container, 'Store Generator'));
+
+		// verify that the element has focus (before changing the value)
+		await waitFor(() => {
+			expect(getByLabelText(container, 'Store Generator')).toHaveFocus();
+		});
+
+		// change the value of the input
+		fireEvent.change(getByLabelText(container, 'Store Generator'), { target: { value: 1 } });
```
JRJurman, line 265: I actually wish we could do a `userEvent` here, but sadly the arrow interactions don't update the input value. If we wanted to make this work we'd have to make a change to the user-events repo - something similar to this: https://github.com/testing-library/user-event/blob/ee062e762f9ac185d982dbf990387e97e05b3c9d/src/pointer/pointerPress.ts#L254-L327

integration-tests/test-app/element-store-generator.ts
```diff
@@ -0,0 +1,34 @@
...
+import elementwithstore from './element-with-store';
+
+const html = registerHtml({
+	elementwithstore,
+});
+
+/**
+ * Element to verify non-standard input controls, and also verify memory leak type issues
+ */
+const elementstoregenerator: TramOneComponent = () => {
```
JRJurman, line 11: See PR summary for the screenshots of this component.

integration-tests/test-app/index.html
```diff
@@ -3,9 +3,9 @@
 	<head>
 		<meta charset="utf-8" />
 		<style>
-			body { background: #0a0f21; color: #cbcbcb; }
-			input { background-color: inherit; color: inherit; }
-			button { background: inherit; color: inherit; }
+			:root {
+				color-scheme: dark light;
+			}
```
JRJurman, line 8: Updated the test-app to use browser native dark / light theme. This should respect the system preferences now!

This is also a change we'd like to make to the generated tram-one apps: https://github.com/Tram-One/tram-one-express/issues/110

integration-tests/test-helpers.ts
```diff
@@ -0,0 +1,19 @@
+import { TramWindow } from '../src/types';
+
+const { waitFor } = require('@testing-library/dom');
+const { startApp } = require('./test-app');
+
+/**
+ * decorated startApp function that ensures that the app's mutation observers
+ * have kicked in before starting to interact with the app
+ */
```
JRJurman, line 9: often times in the unit tests the mutation observers would try to handle multiple interactions at the same time. This did not occur when using the test-app locally, but was a side-effect of jest/testing-library not waiting for the mutation observer to catch up as it was interacting with the app (and, to be fair, rightfully so). 

This decorated `startApp` function starts the app, and then waits for one of our stores (which is updated by the mutation observer) to be populated. This allowed the tests to pass as expected.

src/dom.ts
```diff
@@ -11,7 +11,7 @@ import {
 	restoreWorkingKey,
 } from './working-key';
 import observeTag from './observe-tag';
-import processEffects from './process-effects';
+import processHooks from './process-hooks';
```
JRJurman, line 14: renamed this file / function, which better reflects what it does now (since it processes both effects and state!)

src/effect-store.ts
```diff
@@ -23,8 +23,8 @@ export const {
  * clear the effect store
  * usually called when we want to empty the effect store
  */
-export const clearEffectStore = (effectName: string) => {
-	const effectStore = getEffectStore(effectName);
+export const clearEffectStore = (effectStoreName: string) => {
+	const effectStore = getEffectStore(effectStoreName);
```
JRJurman, line 27: this was unclear, and threw me off, so renamed this to be explicitly the EffectStore's Name.

src/key-store.ts
```diff
@@ -0,0 +1,48 @@
+/*
+ * The KeyStore in Tram-One is a basic key-value object
+ * that needs to be persisted in the globalSpace.
+ *
+ * Currently this is used with useStore and useGlobalStore to keep
+ * track of what stores need to be cleaned up when removing elements
+ */
```
JRJurman, line 7: This is heavily inspired by the EffectStore.

This store keeps track of store keys (e.g. `"app[{}]/test[{}]"`, and how many observers they have (just a number). 

If the number of observers goes down to zero after removing an element from the DOM, then we remove the store.

src/key-queue.ts
```diff
@@ -0,0 +1,31 @@
+/*
+ * The KeyQueue in Tram-One is a basic list of keys
+ * that needs to be persisted in the globalSpace.
+ *
+ * Currently this is used with useStore to keep track of what
+ * stores need to be associated with generated elements
+ */
```
JRJurman, line 7: This KeyQueue (while different functionally) is similar to the EffectQueue. It is used to determine what store keys are generated during element generation, and tie it back to the element. 

The name `KeyQueue` is intentionally generic, as (hopefully) we may consolidate the EffectQueue into this, and the EffectStore into the KeyStore.

src/working-key.ts
```diff
@@ -8,16 +8,17 @@ import { WorkingkeyObject } from './types';
...
-} as WorkingkeyObject;
+const defaultWorkingKey = () =>
+	({
+		// list of custom tags that we've stepped into
+		branch: [],
+		// map of branches to index value (used as a cursor for hooks)
+		branchIndices: {
+			'': 0,
+		},
+	} as WorkingkeyObject);
```
JRJurman, line 19: build a new object (previously we would always point to the same object, which created stale data between tests).

src/process-hooks.ts
```diff
@@ -23,12 +28,19 @@ export default (tagFunction: () => TramOneElement) => {
...
+	// get all new keys
+	const newKeys = getKeyQueue(TRAM_KEY_QUEUE);
+
 	// store new effects in the node we just built
 	const newEffects = Object.keys(queuedEffects).filter((effect) => !(effect in existingEffects));
 	tagResult[TRAM_TAG_NEW_EFFECTS] = newEffects.map((newEffectKey) => queuedEffects[newEffectKey]);
 
-	// restore the effect queue to what it was before we started
+	// store keys in the node we just built
+	tagResult[TRAM_TAG_STORE_KEYS] = newKeys;
```
JRJurman, line 39: this is what associates the observable keys with the element. Basically, after we've generated all the elements (and called any store hooks), we put those keys on the node, which can then be picked up by the mutation observer.

src/observe-tag.ts
```diff
@@ -120,18 +120,27 @@ export default (tagFunction: () => TramOneElement): TramOneElement => {
...
+						elementToGiveFocus.setSelectionRange(
+							removedElementWithFocusData.selectionStart,
+							removedElementWithFocusData.selectionEnd,
+							removedElementWithFocusData.selectionDirection
+						);
+					}
+				} catch (exception) {
+					// don't worry if we fail
+					// this can happen if the element has a `setSelectionRange` but it isn't supported
+					// e.g. input with type="range"
```
JRJurman, line 134: It's slightly baffling that range inputs have a `setSelectionRange` method, but don't actually support it, it's kinda strange, but there's no real way to avoid this other than put a try-catch around it.

JRJurman: We could potentially catch the specific exception, and throw other exceptions we don't expect (although honestly I don't think we want to throw any exception, this is a very non-critical path for the app).

src/observable-hook.ts
```diff
@@ -32,6 +33,13 @@ export default <Store extends StoreObject>(key?: string, value?: Store): Store =
 	// get value for key
 	const keyValue = observableStore[resolvedKey];
 
+	// if we weren't passed in a key, this is a local obserable (not global),
+	const isLocalStore = !key;
+	if (isLocalStore) {
+		// if this is local, we should associate it with the element by putting it in the keyQueue
+		getKeyQueue(TRAM_KEY_QUEUE).push(resolvedKey);
+	}
```
JRJurman, line 41: this is what adds this observable's key to the queue, which is later picked up and added to the element attributes.

src/mutation-observer.ts
```diff
@@ -51,20 +73,48 @@ const processEffects = (node: Node | TramOneElement) => {
...
+ */
+const cleanUpObservableStores = () => {
+	const observableStore = getObservableStore(TRAM_OBSERVABLE_STORE);
+	const keyStore = getKeyStore(TRAM_KEY_STORE);
+	Object.entries(keyStore).forEach(([key, observers]) => {
+		if (observers === 0) {
+			delete observableStore[key];
+		}
+	});
+};
```
JRJurman, line 104: In the above functions we increment or decrement the number of observers associated with a key. If at the end of the day there are zero observers, this is where we delete the store.

src/observe-tag.ts
```diff
@@ -120,18 +120,27 @@ export default (tagFunction: () => TramOneElement): TramOneElement => {
...
 				}
 
 				elementToGiveFocus.scrollLeft = removedElementWithFocusData.scrollLeft;
 				elementToGiveFocus.scrollTop = removedElementWithFocusData.scrollTop;
 			}
 
+			// don't lose track that this is still a tram-one element
+			tagResult[TRAM_TAG] = true;
```
JRJurman, line 142: we totally forgot to do this before :sweat_smile:

src/mutation-observer.ts
```diff
@@ -51,20 +73,48 @@ const processEffects = (node: Node | TramOneElement) => {
...
+/**
+ * remove any stores that no longer have any elements associated with them
+ * see removeStoreKeyAssociation above
+ */
+const cleanUpObservableStores = () => {
+	const observableStore = getObservableStore(TRAM_OBSERVABLE_STORE);
+	const keyStore = getKeyStore(TRAM_KEY_STORE);
+	Object.entries(keyStore).forEach(([key, observers]) => {
+		if (observers === 0) {
+			delete observableStore[key];
```
JRJurman, line 101: we should also delete the key from the keystore (otherwise we'll keep trying to delete these, even when they are already gone).

chtinahow (approved):
LGTM

src/mutation-observer.ts
JRJurman: we now also delete the key from the keyStore

src/key-queue.ts
```diff
@@ -0,0 +1,31 @@
...
+
+/**
+ * clear the key queue
+ * usually called when we want to empty the key queue
+ */
+export const clearKeyQueue = (keyQueueName: string) => {
+	const keyQueue = getKeyQueue(keyQueueName);
+
+	keyQueue.splice(0, keyQueue.length);
+};
```
ethanjurman, line 25: is slice used here to provide a mutation on a reference or could we not do some assignment? Or maybe I'm getting caught up in the naming of this function (this doesn't look feel a true "clear" operation).

ethanjurman (approved):
I have one question but otherwise looks good!

src/key-queue.ts
JRJurman: I believe splice will allow us to empty the array. To be honest I did look this up, and there was also an option to just reassign it to an empty array, but we need to make sure that other references to this variable are cleared out, hence splice. Performance wise it was documented as the same as reassignment

PR #194, opened 2022-06-12
Type Updates - Tram One Generics

JRJurman:
## Summary
Update Tram-One types to use generics for component definitions! 

This mimics React's component prop type definitions.

## Version Update: Patch
This is a patch because it fixes a previously broken workflow. While this technically could introduce typescript errors where there weren't any before, those should be considered broken.

## Example
```ts
type pageHeaderProps = {
  index: number;
};

const pageHeader: TramOneComponent<pageHeaderProps> = ({ index }) => {
  const targetGroupPage = useTargetGroupPage(index);

  return html`
    <h1 class="page-header">
      <span>${targetGroupPage.title || ""}</span>
    </h1>
  `;
```

## Checklist
- [x] PR Summary
- [x] PR Annotations
- [x] ~Tests~
- [x] Version Bump

### DIFF
--- a/integration-tests/test-app/sub-title.ts
+++ b/integration-tests/test-app/sub-title.ts
@@ -5,7 +5,7 @@ const html = registerHtml();
 /**
  * component to test basic rendering
  */
-const subTitle: TramOneComponent = (props: any, children: Element) => {
+const subTitle: TramOneComponent = (props, children) => {
 	return html` <h2>${children}</h2> `;
 };
 
--- a/integration-tests/test-app/title.ts
+++ b/integration-tests/test-app/title.ts
@@ -1,4 +1,4 @@
-import { registerHtml, TramOneComponent, Props } from '../../src/tram-one';
+import { registerHtml, TramOneComponent } from '../../src/tram-one';
 import subtitle from './sub-title';
 
 const html = registerHtml({
@@ -9,7 +9,7 @@ const html = registerHtml({
 /**
  * component to test basic rendering
  */
-const title: TramOneComponent = ({ subtitle }: Props, children: Element) => {
+const title: TramOneComponent = ({ subtitle }, children) => {
 	return html`
 		<header>
 			<h1 class="title">Home Page</h1>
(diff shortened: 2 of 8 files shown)


integration-tests/test-app/sub-title.ts
```diff
@@ -5,7 +5,7 @@ const html = registerHtml();
 /**
  * component to test basic rendering
  */
-const subTitle: TramOneComponent = (props: any, children: Element) => {
```
JRJurman, line 8: this was previously wrong, children can be a list of elements, not a single element

integration-tests/test-app/title.ts
```diff
@@ -9,7 +9,7 @@ const html = registerHtml({
 /**
  * component to test basic rendering
  */
-const title: TramOneComponent = ({ subtitle }: Props, children: Element) => {
```
JRJurman, line 12: same as above, children can be a list of elements, not an Element

package.json
```diff
@@ -1,6 +1,6 @@
 {
 	"name": "tram-one",
-	"version": "12.0.2",
+	"version": "12.0.3",
```
JRJurman, line 3: patch, see PR description for details

src/mount.ts
```diff
@@ -1,12 +1,12 @@
...
-import { Container, TramOneComponent } from './types';
+import { Container, RootTramOneComponent } from './types';
 
 /**
  * Updates a container with an initial component for the first render.
  * @param component the tram-one component to render
  * @param container an element to render the component on
  */
-export default (component: TramOneComponent, container: Container) => {
+export default (component: RootTramOneComponent, container: Container) => {
```
JRJurman, line 9: updated here (and in start.ts) to use RootTramOneComponent - a variant of TramOneComponent with an empty props and no children (technically we could enable the user to pass in props / children in the future - which would be great for server-side rendering, but that's for a future PR)

chtinahow (approved):
Lgtm

JRJurman:
This was merged in as 12.0.2, but is actually 12.0.3 - sorry about that!

PR #195, opened 2022-06-18
Update Dependencies

JRJurman:
## Summary
PR to update the version of `useUrlParams`. Also taking this opportunity to update a bunch of the dev dependencies (most notably `user-events`).

## Version Bump - Major
Since `useUrlParams` had a major bump, this project also inherits that major version bump. In reality, this shouldn't impact most users, but it's still a breaking change.

## Checklist
- [x] PR Summary
- [x] PR Annotations
- [x] Tests
- [x] Version Bump


integration-tests/regression.test.js
```diff
@@ -86,23 +86,23 @@ describe('Tram-One', () => {
 		// focus would be removed from the component and put on the body of the page
 
 		// focus on the input
-		userEvent.click(getByLabelText(container, 'New Task Label'));
+		await userEvent.click(getByLabelText(container, 'New Task Label'));
```
JRJurman, line 89: user-events changed to be a promise API, this might mean we can remove a bunch of the `await waitFor` lines actually!

chtinahow (approved):
👍

Issue #173, opened 2021-08-09
Support components that return an Array of Elements (Fragments)


JRJurman:
## Summary
React supports this using [Fragments](https://reactjs.org/docs/fragments.html), we should look at doing something similar. Maybe [DocumentFragments](https://developer.mozilla.org/en-US/docs/Web/API/DocumentFragment) :man_shrugging: ? This may be unnaturally hard given that we attach so much data to the element being generated :thinking:

JRJurman:
This may be accomplished when we update our dependencies from their original root (see https://github.com/Tram-One/nanohtml/issues/19, https://github.com/Tram-One/hyperx/issues/7, and https://github.com/Tram-One/nanomorph/issues/4)

JRJurman:
Some of this work has been done here: https://github.com/Tram-One/tram-one/commit/b40f101cec87d29d747b35492e393ee2b12d2c79

JRJurman:
We implemented fragments in #197 , if there is a strong case, we could look into supporting returning a list of elements, and wrapping that in an implicit fragment, but that would likely over-complicate the type interface.

PR #196, opened 2022-06-26
Existing Effects & Refs

### LINKED ISSUE #112: top level component requires <div> tag
In the development of the most recent version of Tram-One, it was found that the app must always have a top-level div for the app to get injected into the page.

This shouldn't be an issue for most apps, but certainly something worth documenting and potentially fixing in a future release.

### LINKED ISSUE #130: implement useRef hook
## Summary

React has a useRef hook which is useful for having a mutable value that persists across renders and doesn't trigger re-renders itself. We should add a similar hook for tram-one!

## Potential Implementation

Similar to how the useState hook works, we should use the workingKey value to create uniquely namespaced keys. However we can create a direct mapping to the global space with this namespaced key (rather than creating a Hover Engine action group).

### LINKED ISSUE #166: Effects don't attach when a component is just rendering another component
## Summary

While a parent component only renders another component, any effects that would be attached to the child are blown away. This is because the element returned here is just the child element, so any effects get overridden.

https://github.com/Tram-One/tram-one/blob/412b7c9836d704c58e4923c2ec85340f66625ffe/src/process-effects.ts#L19-L20

Potentially we can warn when a parent would only render a child, or merge these effects. It's unclear if merging would be safe, and in reality, I think having a parent that just returns a child is fairly rare.

JRJurman:
## Summary
This PR makes updates to how `useEffect` works, mostly in prep for #173 

Fixes #130 
Fixes #166
Fixes #112 

### Changes
* process existing effects 
  * these were prepared a while ago (in #142), but we finally found a way to test them in #166
* add `ref` parameter to `useEffect`.
  * this has been discussed for a while in #130, this solution feels the most directly applicable, in the example applications built so far.  

### Additional Changes
* changed top level element to be `<tram-one>` rather than `<div>`, this is more clear and obvious, and won't cause formatting issues (since `<tram-one>` should be undecorated). 
  * #112
* make the pattern parameter for `useUrlParams` optional (the underlying library has a default value here, so no parameter should be required).
* Some underlying types were more explicitly defined
  * This was required to make dealing with the `ref` parameter in `useEffect` much easier (and not require wrapping in a cast to `as unknown`).

## Version Bump: Minor
This PR adds new functionality, that should not be breaking in any previously declared way. It's not required by any means for existing projects, so minor feels like the most appropriate here.

## Checklist
- [x] PR Summary
- [x] PR Annotations
- [x] Tests
- [x] Version Bump

### DIFF
--- a/integration-tests/test-app/focus-input.ts
+++ b/integration-tests/test-app/focus-input.ts
@@ -7,7 +7,7 @@ const html = registerHtml();
  */
 const focusInput: TramOneComponent = () => {
 	useEffect((ref) => {
-		(ref as unknown as HTMLElement).focus();
+		ref.focus();
 	});
 	return html`<input placeholder="Input for automatic focus" />`;
 };
--- a/integration-tests/test-app/focus-input.ts
+++ b/integration-tests/test-app/focus-input.ts
@@ -5,11 +5,11 @@ const html = registerHtml();
 /**
  * Element to test useEffect reference point
  */
-const tramInput: TramOneComponent = () => {
+const focusInput: TramOneComponent = () => {
 	useEffect((ref) => {
 		(ref as unknown as HTMLElement).focus();
 	});
 	return html`<input placeholder="Input for automatic focus" />`;
 };
 
-export default tramInput;
+export default focusInput;
--- a/src/dom.ts
+++ b/src/dom.ts
@@ -14,7 +14,7 @@ import observeTag from './observe-tag';
 import processHooks from './process-hooks';
 import { TRAM_TAG, TRAM_TAG_NEW_EFFECTS, TRAM_TAG_CLEANUP_EFFECTS } from './node-names';
 
-import { Registry, Props, DOMTaggedTemplateFunction, Children } from './types';
+import { Registry, Props, DOMTaggedTemplateFunction, Children, TramOneHTMLElement, TramOneSVGElement } from './types';
 
 /**
  * This function takes in a namespace and registry of custom components,
@@ -25,7 +25,10 @@ import { Registry, Props, DOMTaggedTemplateFunction, Children } from './types';
  * @param registry mapping of tag names to component functions
  * @param namespace namespace to create nodes in (by default XHTML namespace)
  */
-export const registerDom = (namespace: string | null, registry: Registry = {}): DOMTaggedTemplateFunction => {
+export const registerDom = <ElementType extends TramOneHTMLElement | TramOneSVGElement>(
+	namespace: string | null,
+	registry: Registry = {}
+) => {
 	// modify the registry so that each component function updates the hook working key
 	const hookedRegistry = Object.keys(registry).reduce((newRegistry, tagName) => {
 		const tagFunction = registry[tagName];
@@ -68,5 +71,5 @@ export const registerDom = (namespace: string | null, registry: Registry = {}): 
 		return { ...newRegistry, [tagName]: hookedTagFunction };
 	}, {});
 
-	return rbel(hyperx, nanohtml(namespace), hookedRegistry);
+	return rbel(hyperx, nanohtml(namespace), hookedRegistry) as DOMTaggedTemplateFunction<ElementType>;
 };
--- a/src/dom.ts
+++ b/src/dom.ts
@@ -59,7 +59,8 @@ export const registerDom = (namespace: string | null, registry: Registry = {}): 
 			tagResult[TRAM_TAG] = true;
 			// we won't decorate TRAM_TAG_REACTION, that needs to be done later when we observe the tag
 			tagResult[TRAM_TAG_NEW_EFFECTS] = tagResult[TRAM_TAG_NEW_EFFECTS] || [];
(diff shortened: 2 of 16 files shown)


src/mount.ts
```diff
@@ -13,6 +13,6 @@ export default (component: RootTramOneComponent, container: Container) => {
 
 	// this sadly needs to be wrapped in some element so we can process effects
 	// otherwise the root node will not have effects applied on it
-	const renderedApp = html`<div><app /></div>`;
+	const renderedApp = html`<tram-one><app /></tram-one>`;
```
JRJurman, line 16: In the UI, this appears like the following:
image

This is better than `<div>`, because we won't inherit any unexpected styles or formatting.

src/use-url-params.ts
```diff
@@ -17,7 +17,7 @@ import { UrlMatchResults } from './types';
  *
  * @returns object with a `matches` key, and (if it matched) path and query parameters
  */
-export default (pattern: string): UrlMatchResults => {
+export default (pattern?: string): UrlMatchResults => {
```
JRJurman, line 20: this should have always been optional! This better matches the interface of the underlying library.

src/mutation-observer.ts
```diff
@@ -55,7 +55,7 @@ const processTramTags = (node: Node | TramOneElement) => {
 			const effectReaction = observe(() => {
 				// verify that cleanup is a function before calling it (in case it was a promise)
 				if (typeof cleanup === 'function') cleanup();
-				cleanup = effect();
+				cleanup = effect(node);
```
JRJurman, line 58: this is what is required for allowing us to accept ref as a parameter for `useEffect` - we pass the node we are processing back into the effect it defined!

src/dom.ts
```diff
@@ -59,7 +59,8 @@ export const registerDom = (namespace: string | null, registry: Registry = {}):
 			tagResult[TRAM_TAG] = true;
 			// we won't decorate TRAM_TAG_REACTION, that needs to be done later when we observe the tag
 			tagResult[TRAM_TAG_NEW_EFFECTS] = tagResult[TRAM_TAG_NEW_EFFECTS] || [];
-			tagResult[TRAM_TAG_CLEANUP_EFFECTS] = tagResult[TRAM_TAG_NEW_EFFECTS] || [];
+			// cleanup effects will be populated when new effects are processed
+			tagResult[TRAM_TAG_CLEANUP_EFFECTS] = [];
```
JRJurman, line 63: there's no reason this should have been populated by the existing new effects before...

integration-tests/test-app/focus-input.ts
```diff
@@ -0,0 +1,15 @@
+import { registerHtml, useEffect, TramOneComponent } from '../../src/tram-one';
+
+const html = registerHtml();
+
+/**
+ * Element to test useEffect reference point
+ */
+const focusInput: TramOneComponent = () => {
+	useEffect((ref) => {
+		(ref as unknown as HTMLElement).focus();
```
JRJurman, line 10: frustratingly, since `ref` is of type TramOneElement, there's no guarantee that it's an HTMLElement or SVGElement - so we have to do this casting to unknown. 

I'll look into ways we can maybe avoid this, potentially having generics in the registerDOM parts of the code.

JRJurman: Fixed this by adding a ton of type definitions around the dom-wrappers logic.

src/types.ts
```diff
@@ -108,6 +108,9 @@ export interface TramOneElement extends Element {
 	[TRAM_TAG_STORE_KEYS]: string[];
 }
 
+export type TramOneHTMLElement = TramOneElement & HTMLElement;
+export type TramOneSVGElement = TramOneElement & SVGElement;
+
```
JRJurman, line 113: new type definitions, because TramOneElement is too generic, and in reality (right now) it's always an HTMLElement or SVGElement.

If we ever expose registerDOM, people should be able to make their own TramOneElement definitions (which shouldn't be too hard to do).

src/process-hooks.ts
```diff
@@ -28,15 +28,25 @@ export default (tagFunction: () => TramOneElement) => {
...
+	// pull new effects that have yet to be processed from the tag
+	// these can appear when a component re-exposes another component at its root
+	const existingNewEffects = tagResult[TRAM_TAG_NEW_EFFECTS] || [];
 
-	// store new effects in the node we just built
+	// store new effects (and the existing new effects) in the node we just built
 	const newEffects = Object.keys(queuedEffects).filter((effect) => !(effect in existingEffects));
-	tagResult[TRAM_TAG_NEW_EFFECTS] = newEffects.map((newEffectKey) => queuedEffects[newEffectKey]);
+	const newEffectFunctions = newEffects.map((newEffectKey) => queuedEffects[newEffectKey]);
+	const existingNewAndBrandNewEffects = existingNewEffects.concat(newEffectFunctions);
```
JRJurman, line 38: we should write a test to verify that both the top-level and child components' effects are processed

chtinahow (approved):
Conditional approval based on a test for those merged elements 👍

PR #197, opened 2022-07-01
initial work to support fragments

JRJurman:
## Summary
This PR enables fragment support in Tram-One.

This enables partial component building without cluttering the DOM, and building components that have slots for direct children.

```typescript
import { registerHtml, TramOneComponent } from '../../src/tram-one';

const html = registerHtml();

const composite: TramOneComponent = () => {
  const header = html`
    <>
      <h3>Top of Description</h3>
    </>
  `;
  const body = html`
    <>
      Some Details
    </>
  `;

  return html`
    <section>
      ${header}
      ${body}
    </section>
  `;
};

export default composite;
```

This component results in the following DOM:
```html
<section>
  <h3>Top of Description</h3>
  Some Details
</section>
```

## Checklist
- [x] PR Summary
- [x] PR Annotations
- [x] Tests
- [x] Version Bump


src/dom-wrappers.ts
```diff
@@ -13,7 +13,7 @@ import { Registry, TramOneHTMLElement, TramOneSVGElement } from './types';
  * @return tagged template function that builds HTML components
  */
 export const registerHtml = (registry?: Registry) => {
-	return registerDom<TramOneHTMLElement>(null, registry);
+	return registerDom<TramOneHTMLElement>(null, { ...registry, '': fragment });
```
JRJurman, line 16: we're including fragment as a default element in the registry, this means that (like React) it's included by default!

src/dom.ts
```diff
@@ -48,7 +48,11 @@ export const registerDom = <ElementType extends TramOneHTMLElement | TramOneSVGE
 				// reset working key so we have the correct place when starting a new component
 				restoreWorkingKey(TRAM_HOOK_KEY, uniqueBranch);
 
-				return tagFunction(props, children);
+				const processedChildren: Children = children?.map((child: unknown | Element) => {
+					if (child instanceof Element) return child;
+					return `${child}`;
+				});
+				return tagFunction(props, processedChildren);
```
JRJurman, line 55: previously we weren't stringifying children, and using template variables actually creates non-string children!

e.g. `<div>Hello ${() => {}}</div>` would previously generate a child element that is of type `function`.

Since other parts of the code (and Tram-One in general) expect either a `string` or an `Element`, we force this processing.

src/fragment.ts
```diff
@@ -0,0 +1,21 @@
+import { TramOneComponent } from './types';
+import { registerDom } from './dom';
+import useEffect from './use-effect';
+
+const xml = registerDom(null);
```
JRJurman, line 5: we use `registerDom` here because fragments are universal across HTML and SVG (additionally, the `registerHtml` and `registerSvg` use this component, so we couldn't cyclically depend on it).

src/fragment.ts
```diff
@@ -0,0 +1,21 @@
...
+		(children || []).forEach((child) => {
+			if (typeof child === 'string') {
+				ref.insertAdjacentText('beforebegin', child);
+			} else {
+				ref.insertAdjacentElement('beforebegin', child);
+			}
+		});
+		ref.remove();
+	});
+	return xml`<tram-fragment>${children}</tram-fragment>`;
```
JRJurman, line 18: `<tram-fragment>` isn't a real tag per-se, rather it's just a placeholder that will be immediately removed at the end of the `useEffect`.

ethanjurman (approved):
@JRJurman ✨ LGTM

PR #198, opened 2022-07-05
Add props to TramOneElements for devtools

JRJurman:
## Summary
This PR adds properties to Tram-One elements so that they can expose more details in the Properties sidebar panel in Chrome Devtools. This will eventually power the Tram-One chrome extension, which will look like the following:

image

## Version Bump - Patch
This is just a patch version bump, since it should have no effect on the API, and no significant change in performance / development.

## Checklist
- [x] PR Summary
- [x] PR Annotations
- [x] ~Tests~ (No tests, for now)
- [x] Version Bump

### DIFF
--- a/src/dom.ts
+++ b/src/dom.ts
@@ -12,7 +12,7 @@ import {
 } from './working-key';
 import observeTag from './observe-tag';
 import processHooks from './process-hooks';
-import { TRAM_TAG, TRAM_TAG_NEW_EFFECTS, TRAM_TAG_CLEANUP_EFFECTS } from './node-names';
+import { TRAM_TAG, TRAM_TAG_NEW_EFFECTS, TRAM_TAG_CLEANUP_EFFECTS, TRAM_TAG_NAME, TRAM_TAG_PROPS } from './node-names';
 
 import { Registry, Props, DOMTaggedTemplateFunction, Children, TramOneHTMLElement, TramOneSVGElement } from './types';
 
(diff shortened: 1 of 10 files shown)


src/dom.ts
```diff
@@ -69,6 +76,13 @@ export const registerDom = <ElementType extends TramOneHTMLElement | TramOneSVGE
 			// cleanup effects will be populated when new effects are processed
 			tagResult[TRAM_TAG_CLEANUP_EFFECTS] = [];
 
+			// save properties for development debugging
+			if (process.env.NODE_ENV === 'development') {
+				tagResult[TRAM_TAG_NAME] = tagName;
+				tagResult[TRAM_TAG_PROPS] = props;
+				tagResult[TRAM_TAG_CHILDREN] = children;
+			}
```
JRJurman, line 84: these properties are not otherwise on the tag, so we need to add them (but only for development).

`tagName` here is based on what the parent component used to add this to the registry.

src/start.ts
```diff
@@ -59,6 +60,9 @@ export default (component: RootTramOneComponent, target: ElementOrSelector) => {
 	// setup key queue for new observable stores when resolving mounts
 	setupKeyQueue(TRAM_KEY_QUEUE);
 
+	// setup key queue for global observable stores when resolving mounts
+	setupKeyQueue(TRAM_GLOBAL_KEY_QUEUE);
```
JRJurman, line 64: not a huge fan of setting up a whole new key queue here, but there really isn't a better way to associate hooks used in the render with their elements.

src/types.ts
```diff
@@ -106,6 +110,12 @@ export interface TramOneElement extends Element {
 	[TRAM_TAG_NEW_EFFECTS]: Effect[];
 	[TRAM_TAG_CLEANUP_EFFECTS]: CleanupEffect[];
 	[TRAM_TAG_STORE_KEYS]: string[];
+
+	// development properties, these are not guarnteed to be on the element
+	[TRAM_TAG_NAME]?: string;
+	[TRAM_TAG_PROPS]?: Props;
+	[TRAM_TAG_CHILDREN]?: Children;
+	[TRAM_TAG_GLOBAL_STORE_KEYS]?: string[];
```
JRJurman, line 118: even if the app is running in development mode, some properties (like props or children) could be undefined.

chtinahow (approved):
After doing the code review, LGTM

